PR #3, opened 2020-04-14
Add unit tests for the "happy" paths

ejona86:
Gets us to "28.9% of statements", which is pathetic, but sort of
expected for a command line tool like this. main() and getHostIp() are
entirely uncovered.

-------------------

While everything else in the repo is trivial, I wouldn't be surprised if something in these tests strikes you as wrong/needing improvement. That said, they're still close to trivial.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -16,12 +16,14 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"math/rand"
 	"net"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/google/uuid"
 )
 
@@ -62,7 +64,15 @@ func TestGenerate(t *testing.T) {
 		t.Fatalf("want no error, got: %v", err)
 	}
 	if want != string(got) {
-		t.Fatalf("\nwant:\n%v\n----------------\ngot:\n%v",
+		var wantParsed, gotParsed interface{}
+		err1 := json.Unmarshal([]byte(want), &wantParsed)
+		err2 := json.Unmarshal(got, &gotParsed)
+		if err1 != nil || err2 != nil {
+			t.Logf("problem parsing json, error for want: %v, got: %v", err1, err2)
+		} else if diff := cmp.Diff(wantParsed, gotParsed); diff != "" {
+			t.Fatalf("not equal (-want +got):\n%s", diff)
+		}
+		t.Fatalf("not equal, but structure matched\nwant:\n%v\n----------------\ngot:\n%v",
 			want, string(got))
 	}
 }
--- a/main_test.go
+++ b/main_test.go
@@ -1,0 +1,120 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


ejona86:
@dfawley, i can haz test?

(Or in other words: Esteemed Mr. Fawley, You are cordially invited to participate in a code viewing ceremony at #3 traffic-director-grpc-bootstrap, GoogleCloudPlatform, GitHub Inc on the week of May 11, 2020. Refreshments will be served, barring a global pandemic.)

main_test.go
```diff
@@ -0,0 +1,120 @@
...
+      "zone": "uscentral-5"
+    }
+  }
+}`
+	out, err := generate(in)
+	if err != nil {
+		t.Fatalf("expected no error, got: %v", err)
+	}
+	if expected != string(out) {
+		t.Fatalf("\nexpected:\n%v\n----------------\ngot:\n%v",
```
dfawley, line 65: This is going to be hard to make sense of if it fails.

One option would be to do a text `diff`, but it seems there is no canonical package for that.  Another option is to JSON parse and use https://godoc.org/github.com/google/go-cmp/cmp#Diff.  Yet another option would be to make sure the test never fails.

main_test.go
```diff
@@ -0,0 +1,120 @@
...
+			expected, string(out))
+	}
+}
+
+func TestGetZone(t *testing.T) {
+	server := httptest.NewServer(nil)
+	defer server.Close()
+	overrideHttp(server)
```
dfawley, line 73: Optional: put the `NewServer` call in `overrideHttp` and have that return the server.

dfawley, line 73: Go style: `overrideHTTP`

main_test.go
```diff
@@ -0,0 +1,120 @@
...
+func TestGenerate(t *testing.T) {
+	uuid.SetRand(rand.New(rand.NewSource(1)))
+	in := configInput{
+		xdsServerUri:     "example.com:443",
+		gcpProjectNumber: 123456789012345,
+		vpcNetworkName:   "thedefault",
+		ip:               "10.9.8.7",
+		zone:             "uscentral-5",
+	}
+	expected := `{
```
dfawley, line 37: Go convention is "got" and "want" instead of "out"/"actual" and "expected".

main_test.go
ejona86: I considered that and then had a reason not to do it. I think it was for clarity. I see the `nil` passed to NewServer is important, because it means we're using the default mux. That could be an argument to overrideHTTP() though. I think it was also that I wanted the server.close() to be more obvious; it isn't as obvious it needs to be shut down as a return value of overrideHTTP().

main_test.go
ejona86: I know it is hard to read. Initially I didn't set the random seed for UUID. It took me quite a bit of time to track down the problem to the UUID changing (it was the first time it came close to looking correct, so I assumed indentation was wrong or some such, because I had copied-and-pasted from a previous test run's output). I poked around and didn't see many alternatives.

I like the idea of cmp.Diff, although if it finds them to be equal and but the raw string is different I think I'll still output this hard-to-read version. I'll fix that up tomorrow.

ejona86:
@dfawley, friendly ping

main_test.go
dfawley: It's `nil` in both places it is called, so the parameter doesn't actually seem all that important to the test as-is.  I'm fine with this either way, though.

dfawley (approved):

ejona86:
I'm going to prefer explicit over implicit for the server thing. It is easy to change in the future if it becomes annoying.

PR #5, opened 2020-11-06
Emit file_watcher plugin configuration with default values

easwars:
This change makes the bootstrap generator emit the file_watcher plugin configuration by default. Currently there are no flags added to control this behavior. I can add them if we think it is necessary.

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,6 +52,7 @@ func main() {
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Error: failed to determine project id: %s\n", err)
 			os.Exit(1)
+
 		}
 	}
 	ip, err := getHostIp()
--- a/main.go
+++ b/main.go
@@ -35,14 +35,9 @@ var (
 	outputName       = flag.String("output", "-", "output file name")
 	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName = flag.String("vpc-network-name", "default", "VPC network name")
-
-	// Notice: This flag is EXPERIMENTAL and may be changed or removed in a later release.
-	enableFileWatcherConfig = flag.Bool("enable-file-watcher-config", false, "whether or not to generate file_watcher certificate provider config")
+	vpcNetworkName          = flag.String("vpc-network-name", "default", "VPC network name")
+	enableFileWatcherConfig = flag.Bool("enable-file-watcher-config", false, "whether or not to generate file_watcher certificate provider config. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 )
-
-// For overriding in unit tests.
-var newUUIDString = func() string { return uuid.New().String() }
 
 func main() {
 	flag.Parse()
@@ -52,7 +47,6 @@ func main() {
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Error: failed to determine project id: %s\n", err)
 			os.Exit(1)
-
 		}
 	}
 	ip, err := getHostIp()
@@ -124,7 +118,7 @@ func generate(in configInput) ([]byte, error) {
 			},
 		},
 		Node: &node{
-			Id:      newUUIDString() + "~" + in.ip,
+			Id:      uuid.New().String() + "~" + in.ip,
 			Cluster: "cluster", // unused by TD
 			Locality: &locality{
 				Zone: in.zone,
--- a/main.go
+++ b/main.go
@@ -35,7 +35,9 @@ var (
 	outputName       = flag.String("output", "-", "output file name")
 	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName          = flag.String("vpc-network-name", "default", "VPC network name")
+	vpcNetworkName = flag.String("vpc-network-name", "default", "VPC network name")
+
+	// Notice: This flag is EXPERIMENTAL and may be changed or removed in a later release.
 	enableFileWatcherConfig = flag.Bool("enable-file-watcher-config", false, "whether or not to generate file_watcher certificate provider config")
 )
 
--- a/main.go
+++ b/main.go
@@ -36,17 +36,6 @@ var gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 	"the gcp project number. If unknown, can be found via 'gcloud projects list'")
 var vpcNetworkName = flag.String("vpc-network-name", "default", "VPC network name")
 
-const (
-	privateSPIFFECertProviderInstanceName = "google_cloud_private_spiffe"
-	privateSPIFFECertProviderName         = "file_watcher"
-	privateSPIFFECertificateFile          = "/var/run/gke-spiffe/certs/certificates.pem"
-	privateSPIFFEKeyFile                  = "/var/run/gke-spiffe/certs/private_key.pem"
-	privateSPIFFECAFile                   = "/var/run/gke-spiffe/certs/ca_certificates.pem"
-	// The file_watcher plugin will parse this a Duration proto, but it is totally
-	// fine to just emit a string here.
-	privateSPIFFERefreshInterval = "10m"
-)
-
 // For overriding in unit tests.
 var newUUIDString = func() string { return uuid.New().String() }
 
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -36,6 +36,20 @@ var gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 	"the gcp project number. If unknown, can be found via 'gcloud projects list'")
 var vpcNetworkName = flag.String("vpc-network-name", "default", "VPC network name")
 
+const (
+	privateSPIFFECertProviderInstanceName = "google_cloud_private_spiffe"
```
ejona86, line 40: Any literal that is only used once, please inline it. This indirection just makes things harder to read.

main_test.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2020 Google LLC
```
ejona86, line 1: We need to get https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/pull/3 merged, before this.

ejona86 (commented):
This is an alpha feature, it must be flag protected before it is stable.

main_test.go
easwars: Ack.

main.go

easwars (commented):
PTAL.

main.go
```diff
@@ -30,11 +30,17 @@ import (
...
-var gcpProjectNumber = flag.Int64("gcp-project-number", 0,
-	"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-var vpcNetworkName = flag.String("vpc-network-name", "default", "VPC network name")
+var (
+	xdsServerUri     = flag.String("xds-server-uri", "trafficdirector.googleapis.com:443", "override of server uri, for testing")
+	outputName       = flag.String("output", "-", "output file name")
+	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
+		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
+	vpcNetworkName          = flag.String("vpc-network-name", "default", "VPC network name")
+	enableFileWatcherConfig = flag.Bool("enable-file-watcher-config", false, "whether or not to generate file_watcher certificate provider config")
```
ejona86, line 39: I think we should add EXPERIMENTAL or something to the description. We'll want to remove this flag in the future.

ejona86 (commented):
The core code looks good and fine. Let's see if we can get #3 merged, and then we can see where to take the tests on this one.

main.go
```diff
@@ -124,6 +132,22 @@ func generate(in configInput) ([]byte, error) {
...
+		c.CertificateProviders = map[string]certificateProviderConfig{
+			"google_cloud_private_spiffe": {
+				PluginName: "file_watcher",
+				Config: privateSPIFFEConfig{
+					CertificateFile:   "/var/run/gke-spiffe/certs/certificates.pem",
+					PrivateKeyFile:    "/var/run/gke-spiffe/certs/private_key.pem",
+					CACertificateFile: "/var/run/gke-spiffe/certs/ca_certificates.pem",
+					// The file_watcher plugin will parse this a Duration proto, but it is totally
+					// fine to just emit a string here.
+					RefreshInterval: "10m",
```
sanjaypujare, line 145: Can we generate "600s" instead of "10m"? Just realized the Java [`Durations.parse(String)`](https://github.com/protocolbuffers/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/Durations.java#L241) only supports "s" as the suffix :-(

ejona86: That should be true for all languages, as that's the protobuf Duration JSON format https://github.com/protocolbuffers/protobuf/blob/0bb6dd8fd77282cdc04c26fa01dbddaec71ac580/src/google/protobuf/duration.proto#L92 .

sanjaypujare: If that's true we should not bother with this Duration stuff at all. Just make it an int and call it `refresh_interval_secs` to make it easy. We thought it was possible to use various suffixes like "m", "d" for minutes or days etc but that's not there.

ejona86: No, we've consistently used Duration for durations in our JSON. Doing something else just adds confusion.

sanjaypujare (changes requested):
We cannot use "10m" as it is not parsable by `Durations.parse()`.

main.go
easwars: Done.

main.go
easwars: Fixed it. Sorry for the confusion. I don't know why I assumed it would support `m`, `d`, `h` etc.

ejona86:
@easwars, #3 is merged.

easwars:
> @easwars, #3 is merged.

Updated after #3. PTAL. Thanks.

main.go
ejona86: We need experimental where the user would see. This comment is useless. I was suggesting in the description string.

main.go
```diff
@@ -44,6 +52,7 @@ func main() {
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Error: failed to determine project id: %s\n", err)
 			os.Exit(1)
+
```
ejona86, line 55: Remove?

main.go
```diff
@@ -30,11 +30,19 @@ import (
...
+	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
+		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
+	vpcNetworkName = flag.String("vpc-network-name", "default", "VPC network name")
+
+	// Notice: This flag is EXPERIMENTAL and may be changed or removed in a later release.
+	enableFileWatcherConfig = flag.Bool("enable-file-watcher-config", false, "whether or not to generate file_watcher certificate provider config")
+)
+
+// For overriding in unit tests.
+var newUUIDString = func() string { return uuid.New().String() }
```
ejona86, line 45: Could we use the fixed seed instead? That way, we actually see that, yes, it is a UUID.

main_test.go
```diff
@@ -17,63 +17,117 @@ package main
...
 func TestGenerate(t *testing.T) {
-	uuid.SetRand(rand.New(rand.NewSource(1)))
-	in := configInput{
-		xdsServerUri:     "example.com:443",
-		gcpProjectNumber: 123456789012345,
-		vpcNetworkName:   "thedefault",
-		ip:               "10.9.8.7",
-		zone:             "uscentral-5",
-	}
-	want := `{
```
ejona86, line 39: I want a test that checks the literal byte output, like was done previously. Otherwise there is no verification for the JSON key names and such. I don't actually see much value of using the parsed version instead of the string, as it is actually more verbose.

easwars: Done.

main.go
easwars: Done.

easwars (commented):
PTAL.

ejona86:
@easwars, push the changes?

easwars:
> @easwars, push the changes?

facepalm
done now

ejona86 (approved):

main.go
```diff
@@ -30,11 +30,14 @@ import (
...
-var gcpProjectNumber = flag.Int64("gcp-project-number", 0,
-	"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-var vpcNetworkName = flag.String("vpc-network-name", "default", "VPC network name")
+var (
+	xdsServerUri     = flag.String("xds-server-uri", "trafficdirector.googleapis.com:443", "override of server uri, for testing")
+	outputName       = flag.String("output", "-", "output file name")
+	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
+		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
+	vpcNetworkName          = flag.String("vpc-network-name", "default", "VPC network name")
+	enableFileWatcherConfig = flag.Bool("enable-file-watcher-config", false, "whether or not to generate file_watcher certificate provider config. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
ejona86, line 39: How did you choose this argument name? I don't see it in the user guide. It seems the name and configuration don't entirely match. It isn't just "file watcher" config. You have to know it is for google_cloud_private_spiffe

easwars: I just came up with something for the flag name. Didn't realize it was going to be specified in the UG.

As for the name and configuration, I think the UG might need to be updated. What is found here should match the config mentioned in Sanjay's latest doc on file watcher plugin.

easwars: I don't know why I can't @ mention Sanjay here.

ejona86: You can, it just doesn't auto-complete since he isn't a member of the repository. Calling @sanjaypujare to coordinate this flag name.

sanjaypujare: So only the flag name is being discussed here, right? The generated config should match the following:
```
 "certificate_providers": {
    "google_cloud_private_spiffe": {
        "plugin_name": "file_watcher",
        "config": {
           "certificate_file": "/var/run/gke-spiffe/certs/certificates.pem",
           "private_key_file": "/var/run/gke-spiffe/certs/private_key.pem",
           "ca_certificate_file": "/var/run/gke-spiffe/certs/ca_certificates.pem",
           "refresh_interval": "600s"
        }
    }
  }
```
The UG needs to be updated with some of this latest info. As far as the flag name is concerned I don't know where in the UG it will be mentioned. If people don't like `enable-file-watcher-config` I can suggest `enable-psm-security` which is shorter and also doesn't depend on file-watcher etc.

easwars: Thank Sanjay.

@ejona86 : Would you prefer the flag name changed to `enable-psm-security`?

ejona86: `--enable-psm-security` sounds good, as it is product-focused. I do think we should probably avoid "enable" though, since it doesn't actually enable any security itself. Maybe `--allow-psm-security` or `--include-psm-security`?

sanjaypujare: I like `--include-psm-security` as it's the most accurate in terms of what the flag is achieving. Let's do that

easwars: Done. Thanks.

PR #6, opened 2020-11-11
Add support for "grpc_server_resource_name_id" field

easwars:
This value will be used as the `id` in the `resource_names` field of the LDS request, with `context_params` containing the listening IP:Port.

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,8 +35,9 @@ var (
 	outputName       = flag.String("output", "-", "output file name")
 	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName     = flag.String("vpc-network-name", "default", "VPC network name")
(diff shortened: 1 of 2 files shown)


ejona86 (commented):
This is not GA-ready. This needs to be flag protected.

main.go
```diff
@@ -128,6 +128,7 @@ func generate(in configInput) ([]byte, error) {
 				"TRAFFICDIRECTOR_GCP_PROJECT_NUMBER": strconv.FormatInt(in.gcpProjectNumber, 10),
 			},
 		},
+		GRPCServerResourceNameId: "grpc/server",
```
ejona86, line 131: FWIW, I thought this was going to be mirroring the udpa resource name format, with a repeated string field.

easwars: It was recently changed to be just a string instead of a list of strings. I'll try to look up where this is documented.

easwars (commented):
> This is not GA-ready. This needs to be flag protected.

Done.

ejona86 (approved):

PR #7, opened 2020-11-11
Update Node proto to work with v2 and v3 xDS implementations in TD.

easwars:
We are interested in the following pieces of information in the `Node` proto:
- network name
- project number
- local IP
- a unique node id

Today (for v2 implementation in TD), these are conveyed as follows:
- local IP and unique node id are encoded in `node.id`
- network and project are sent as `node.metadata` entries

Going forward, we want to do the following:
- Pass network, project and unique node id in the `node.id` field in the format: `projects/{project number}/networks/{network name}/nodes/{id}`
- also continue to send network and project as `node.metadata` entries
- send local IP in `network.metadata` with key `INSTANCE_IP`

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,9 +35,8 @@ var (
 	outputName       = flag.String("output", "-", "output file name")
 	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName          = flag.String("vpc-network-name", "default", "VPC network name")
-	includePSMSecurity      = flag.Bool("include-psm-security", false, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	includeServerResourceID = flag.Bool("include-server-resource-id", false, "whether or not to generate config for server resource id. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	vpcNetworkName    = flag.String("vpc-network-name", "default", "VPC network name")
+	includeV3Features = flag.Bool("include-v3-features", false, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 )
 
 func main() {
@@ -61,13 +60,12 @@ func main() {
 		zone = ""
 	}
 	config, err := generate(configInput{
-		xdsServerUri:            *xdsServerUri,
-		gcpProjectNumber:        *gcpProjectNumber,
-		vpcNetworkName:          *vpcNetworkName,
-		ip:                      ip,
-		zone:                    zone,
-		includePSMSecurity:      *includePSMSecurity,
-		includeServerResourceID: *includeServerResourceID,
+		xdsServerUri:      *xdsServerUri,
+		gcpProjectNumber:  *gcpProjectNumber,
+		vpcNetworkName:    *vpcNetworkName,
+		ip:                ip,
+		zone:              zone,
+		includeV3Features: *includeV3Features,
 	})
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to generate config: %s\n", err)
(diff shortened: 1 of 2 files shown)


easwars:
I've specifically not protected this with a flag as the format change works with the current TD v2 implementation in TD as well.

main.go
```diff
@@ -120,15 +124,20 @@ func generate(in configInput) ([]byte, error) {
...
-			Id:      uuid.New().String() + "~" + in.ip,
+			Id:      fmt.Sprintf(nodeIDFormatStr, in.gcpProjectNumber, in.vpcNetworkName, uuid.New().String()),
 			Cluster: "cluster", // unused by TD
 			Locality: &locality{
 				Zone: in.zone,
 			},
 			Metadata: map[string]string{
 				"TRAFFICDIRECTOR_NETWORK_NAME":       in.vpcNetworkName,
 				"TRAFFICDIRECTOR_GCP_PROJECT_NUMBER": strconv.FormatInt(in.gcpProjectNumber, 10),
+				"INSTANCE_IP":                        in.ip,
```
ejona86, line 140: Where did this key name come from?

main.go
```diff
@@ -40,6 +40,10 @@ var (
 	includeServerResourceID = flag.Bool("include-server-resource-id", false, "whether or not to generate config for server resource id. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 )
 
+// The `id` field of the Node proto is set to a value of the format
+// `projects/{project number}/networks/{network name}/nodes/{uuid}`.
+const nodeIDFormatStr = "projects/%d/networks/%s/nodes/%s"
```
ejona86, line 45: Inline. Splitting a format string from its arguments is asking for trouble.

ejona86 (commented):
I thought this was supposed to happen as we enable v3.

main.go

main.go
easwars: This is mentioned in the Security Alpha User Guide.

easwars:
> I thought this was supposed to happen as we enable v3.

v3 support is now in staging, and we are starting some interop testing.

ejona86:
"starting interop" does not mean it is stable. Something should only be non-flag protected when it is stable. (If we were going to do final testing with the bootstrap change as part of its release process, that could potentially be fine. But saying, "we're still working on the feature" is a void argument for saying why something doesn't need a flag.)

easwars:
This is supposed to work with the existing implementation because we are not changing the way the `project` and `network` are carried. Also, TD already supports getting the IP from the `INSTANCE_IP` metadata key because that is what envoy does I believe.

ejona86:
But does it work in the new implementation? If we had a typo here, or the server actually expected a different format, would we know?

All discussion sounded like this change was to be made when we enabled v3 support. If that's not the case, that's fine, and TD can decide how it goes. But we shouldn't assume otherwise.

easwars:
Ok, let me add a flag :)
I've been trying (and failing) to see whether I can get by without one.

ejona86:
The configuration being written in a doc somewhere says nothing as to whether it is ready. We need to run actual interop tests on production code (or we ran it on dev instances and the server team is rolling it out to production). If we haven't actually tested the system using the configuration to show that it is working as expected, then it needs a flag. In this case, yes v2 shouldn't care. But we are making the change for v3, so unless we have confirmed it works for v3, we shouldn't roll out the change.

ejona86:
@easwars, if you want to just have a "v3" flag, and include this as part of it, that sounds fine to me.

main.go
```diff
@@ -238,11 +241,11 @@ type creds struct {
 }
 
 type node struct {
-	Id           string      `json:"id,omitempty"`
-	Cluster      string      `json:"cluster,omitempty"`
-	Metadata     interface{} `json:"metadata,omitempty"`
```
easwars, line 243: Was there any particular reason to have this as an `interface{}` instead of a concrete type. I just changed it to make it easier to directly add to the map in the `v3` case.

easwars:
@ejona86 : Is this blocked on any changes from my end?

main.go
```diff
@@ -35,9 +35,8 @@ var (
 	outputName       = flag.String("output", "-", "output file name")
 	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName          = flag.String("vpc-network-name", "default", "VPC network name")
-	includePSMSecurity      = flag.Bool("include-psm-security", false, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
ejona86, line 39: PSM security is different than V3 features. V3 will very likely go GA before PSM security.

main.go
ejona86: I don't expect there was any particular reason, other than it was a bit unclear if we'd use a struct or a map.

main.go
easwars: Split them up.

ejona86 (approved):

PR #9, opened 2020-11-30
Add a command line arg for specifying zone.

easwars:


ejona86 (approved):
You may want to delay merging until https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/pull/10 is in, so we can see that it works.

ejona86:
@easwars, #10 worked well. You have write access now, so you can merge and tar.gz and container will automatically be made.

PR #10, opened 2020-11-30
cloudbuild: Build docker image as well

ejona86:
@easwars to review, once he accepts the repository's invite.

CC @sanjaypujare @sergiitk


sergiitk (approved):
At some point we should consider prepending commit sha with date/time to make image tags naturally sortable.

ejona86:
Having date/time also makes them harder to find problematically. I think most UIs for this sort of thing have a sort by newest. But really, you should just be choosing the commit you want and use it.

In some cases we may want to use `latest`, although we need to be careful about that; I don't really trust us using that tag.

easwars (approved):

PR #11, opened 2020-12-02
Add support for ECS metadata labels.

easwars:


ejona86:
Can we please not mix the bug fix (server_features) with a new feature (metadata)?

easwars:
> Can we please not mix the bug fix (server_features) with a new feature (metadata)?

Was taking some liberties since you made me a contributor on this repo :P 
Reverted the other changes now.

ejona86:
Sanjay is going to discuss with some others to determine the requirements for this.

ejona86:
The verdict from Sanjay's meeting is we don't need to manage these as a single string. Let's do the "pass the flag multiple times" approach and get rid of the comma separated stuff. I think we'll also want to give it a more generic name like `--metadata` or the like, and not assume it is tied to ECS.

easwars:
> The verdict from Sanjay's meeting is we don't need to manage these as a single string. Let's do the "pass the flag multiple times" approach and get rid of the comma separated stuff. I think we'll also want to give it a more generic name like `--metadata` or the like, and not assume it is tied to ECS.

Thanks. Done. PTAL.

main.go
```diff
@@ -73,6 +77,7 @@ func main() {
 		zone:               zone,
 		includeV3Features:  *includeV3Features,
 		includePSMSecurity: *includePSMSecurity,
+		ecsMetadataLabels:  nodeMetadata,
```
ejona86, line 80: Remove the "ecs" references.

main.go
```diff
@@ -234,6 +243,24 @@ func getFromMetadata(urlStr string) ([]byte, error) {
 	return body, nil
 }
 
+// Parse the comma separated list of labels specified on the command line as
+// part of the --ecs-metadata-labels flag into a map of key value pairs.
+func parseMetadataLabels(labels string) (map[string]string, error) {
```
ejona86, line 248: Delete?

map_flag.go
```diff
@@ -0,0 +1,61 @@
...
+			b.WriteRune(',')
+		}
+		b.WriteString(key)
+		b.WriteRune('=')
+		b.WriteString((*s)[key])
+	}
+	return b.String()
+}
+
+func (s *stringMapVal) Get() interface{} {
```
ejona86, line 59: What uses this?

main.go

main.go
easwars: Done. Thanks.

map_flag.go
easwars: Looks like I can get rid of it. The `flag` package says that every `Value` type exported by it implements this `Getter` interface. But I think it is mostly for use with `flag.Lookup()`. I don't think we would need that. So, got rid of it.

ejona86 (approved):

PR #14, opened 2021-02-24
Support server_listener_resource_name_template

easwars:


easwars:
I went with the assumption that since this bootstrap generator is meant to be used with TD as the control plane, we can hardcode the listener_name template and not give the option of overriding with a flag. Please let me know if you think otherwise.

main.go
```diff
@@ -243,10 +243,10 @@ func getFromMetadata(urlStr string) ([]byte, error) {
 }
 
 type config struct {
-	XdsServers               []server                             `json:"xds_servers,omitempty"`
-	Node                     *node                                `json:"node,omitempty"`
-	CertificateProviders     map[string]certificateProviderConfig `json:"certificate_providers,omitempty"`
-	GRPCServerResourceNameID string                               `json:"grpc_server_resource_name_id,omitempty"`
+	XdsServers                         []server                             `json:"xds_servers,omitempty"`
+	Node                               *node                                `json:"node,omitempty"`
+	CertificateProviders               map[string]certificateProviderConfig `json:"certificate_providers,omitempty"`
```
sanjaypujare, line 248: The above 3 lines don't contain an actual change - just tab/whitespace changes?

easwars: That's `gofmt`s work.

easwars:
Gentle ping ...

ejona86 (approved):
Since this changes all-at-once, we'll probably need to upgrade this in the test environment at the same time that gRPC java is upgraded.

easwars:
> Since this changes all-at-once, we'll probably need to upgrade this in the test environment at the same time that gRPC java is upgraded.

@sanjaypujare FYI

PR #15, opened 2021-03-19
Emit v3 configs by default

easwars:
Also provide a flag to fallback to v2 configs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,7 +37,7 @@ var (
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName     = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone       = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
-	includeV3Features  = flag.Bool("include-v3-features-experimental", false, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	includeV3Features  = flag.Bool("include-v3-features-experimental", true, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -37,10 +37,12 @@ var (
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName     = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone       = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
-	includeV3Features  = flag.Bool("include-v3-features-experimental", false, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
ejona86, line 40: Leave the old flag and just change the default value. That prevents needing to do an atomic upgrade (e.g., to upgrade the current v3 tests), and is easier on us poor humans.

main.go
```diff
@@ -128,34 +130,47 @@ func generate(in configInput) ([]byte, error) {
...
+	// in TD when this option is set.
+	if in.excludeV3Features {
 		// xDS v2 implementation in TD expects the IP address to be encoded in the
-		// id field while the v3 implementation expects this in the metadata.
-		c.Node.Metadata["INSTANCE_IP"] = in.ip
-		c.XdsServers[0].ServerFeatures = append(c.XdsServers[0].ServerFeatures, "xds_v3")
+		// id field.
+		c.Node.Id = uuid.New().String() + "~" + in.ip
+
+		// Remove the xds_v3 server feature.
```
ejona86, line 164: I don't see why the old code was rewritten. Only adding xds_v3 conditionally seems superior. I think this entire change can be simplified to just changing the default value of the flag to `true`.

main.go
easwars: Done. Thanks for the suggestion.

main.go

ejona86 (approved):
It seems strange to reorder the tests; I'd have expected to just change the desc string. But whatever.

PR #16, opened 2021-03-30
Make directory containing secrets configurable

easwars:
This is required because Zatar allows the user to mount the volume with secrets at a location of their liking.

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


easwars:
cc @sanjaypujare

easwars:
> Looks like this actually changes the default. That is generally appropriate to point out.

That's right. Zatar changed their default as part of their public preview. 

> Can we easily have two flag names for the same variable? When this stabilizes we'll need to change the flag name and that will break existing callers of secrets-dir-experimental

I think we can have two flag names for the same variable. My initial reading for https://golang.org/pkg/flag/#StringVar makes me think so.
But we will also be removing other `--xxx-experimental` flags over time, right? So, why is this flag any different?

ejona86 (approved):

easwars:
I tried experimenting with two flag names for the same variable. It works. The only catch is that whichever flag is specified last on the command line, the variable will take that value (which seems reasonable, but maybe we might have to document that when we get there).

ejona86:
> The only catch is that whichever flag is specified last on the command line, the variable will take that value (which seems reasonable, but maybe we might have to document that when we get there).

That's very fair. I doubt I'll care enough to document it, but either way we are fine. Thanks for checking.

PR #20, opened 2021-08-11
Add deployment info as part of Node metadata

easwars:
Summary of changes
- Identify the deployment type by talking to the Metadata server
- Add command line flags for deployment information that we plan to add to the `Node` metadata
- Talk to Metadata server and retrieve deployment information when the same is not specified on the command line

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,11 +36,16 @@ var (
 	outputName       = flag.String("output", "-", "output file name")
 	gcpProjectNumber = flag.Int64("gcp-project-number", 0,
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName     = flag.String("vpc-network-name", "default", "VPC network name")
-	localityZone       = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
-	includeV3Features  = flag.Bool("include-v3-features-experimental", true, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	includePSMSecurity = flag.Bool("include-psm-security-experimental", false, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	secretsDir         = flag.String("secrets-dir-experimental", "/var/run/secrets/workload-spiffe-credentials", "path to a directory containing TLS certificates and keys required for PSM security. Used only if --include-psm-security-experimental is set. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	vpcNetworkName        = flag.String("vpc-network-name", "default", "VPC network name")
+	localityZone          = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
+	includeV3Features     = flag.Bool("include-v3-features-experimental", true, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	includePSMSecurity    = flag.Bool("include-psm-security-experimental", false, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -36,11 +36,16 @@ var (
...
+	includeV3Features     = flag.Bool("include-v3-features-experimental", true, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	includePSMSecurity    = flag.Bool("include-psm-security-experimental", false, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	secretsDir            = flag.String("secrets-dir-experimental", "/var/run/secrets/workload-spiffe-credentials", "path to a directory containing TLS certificates and keys required for PSM security. Used only if --include-psm-security-experimental is set. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	includeDeploymentInfo = flag.Bool("include-deployment-info-experimental", false, "whether or not to generate config which contains deployment related information. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	gkeClusterName        = flag.String("gke-cluster-name", "", "GKE cluster name to use, instead of retrieving it from the metadata server")
```
ejona86, line 45: experimental?

main.go
```diff
@@ -219,6 +265,41 @@ func getProjectId() (int64, error) {
...
+	return string(cluster)
+}
+
+// For overriding in unit tests.
+var readHostNameFile = func() ([]byte, error) {
+	return ioutil.ReadFile("/etc/hostname")
+}
+
+func getPodName() string {
+	if pod := os.Getenv("HOSTNAME"); pod != "" {
```
ejona86, line 283: Why not `os.Hostname()`?

main.go
easwars: Would it be better to add something like this to the flag description of all these new flags.
```
Only used when --include-deployment-info-experimental is set.
```

What do you think?

main.go
easwars: Yes, that's simpler. Done.

main.go
ejona86: That could be fine.

main.go
```diff
@@ -70,6 +75,42 @@ func main() {
...
+				"GKE-POD":       pod,
+				"GKE-NAMESPACE": *gkeNamespace,
+			}
+		case deploymentTypeGCE:
+			vmName := *gcpVM
+			if vmName == "" {
+				vmName = getVMName()
+			}
+			deploymentInfo = map[string]string{
+				"GCP-VM":      vmName,
```
karthikbox, line 107: nit: can this be GCE-VM to be inline with other places where we use GKE-*

main_test.go
```diff
@@ -157,6 +157,60 @@ func TestGenerate(t *testing.T) {
...
+				ip:                "10.9.8.7",
+				zone:              "uscentral-5",
+				includeV3Features: true,
+				deploymentInfo: map[string]string{
+					"GCP-ZONE":      "uscentral-5",
+					"GKE-CLUSTER":   "test-gke-cluster",
+					"GKE-NAMESPACE": "test-gke-namespace",
+					"GKE-POD":       "test-gke-pod",
+					"INSTANCE-IP":   "10.9.8.7",
+					"GKE-VM":        "test-gce-vm",
```
karthikbox, line 177: QQ: Is GKE-VM this a typo? In main.go, i only see GCP-VM (i left a comment about changing it to GCE-VM). 
How did this end up in node metadata, event though "GKE-VM" is unknown to main.go?

main.go
easwars: Thought about it a little more, and decided to add the `experimental` suffix to the flags. Thought that would be better instead of coupling these flags with the `--include-deployment-info-experimental` flag.

main.go
easwars: Thanks for the catch. Fixed it.

main_test.go
easwars: So, this test only tests the `generate` function. It doesn't test thing end to end. But I've fixed it here and in main.go to use `GCE-VM` instead. Thanks.

ejona86 (approved):

PR #22, opened 2021-09-17
Add --config-scope-experimental flag

gnossen:
The AppNet APIs introduce a `scope` field independent of VPC network. This PR introduces a flag enabling the bootstrap file to select a particular scope.

### DIFF
--- a/main.go
+++ b/main.go
@@ -112,7 +112,7 @@ func main() {
 		}
 	}
 
-	config, err := generate(configInput{
+	input := configInput{
 		xdsServerUri:       *xdsServerUri,
 		gcpProjectNumber:   *gcpProjectNumber,
 		vpcNetworkName:     *vpcNetworkName,
@@ -124,7 +124,16 @@ func main() {
 		metadataLabels:     nodeMetadata,
 		deploymentInfo:     deploymentInfo,
 		configScope:        *configScope,
-	})
+	}
+
+	validationErr := validate(input)
+
+	if validationErr != nil {
+		fmt.Fprintf(os.Stderr, "Error: %s\n", validationErr)
+		os.Exit(1)
+	}
+
+	config, err := generate(input)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to generate config: %s\n", err)
 		os.Exit(1)
@@ -168,6 +177,36 @@ type configInput struct {
 	metadataLabels     map[string]string
 	deploymentInfo     map[string]string
 	configScope        string
+}
+
+func isLetter(b byte) bool {
+	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z')
+}
+
+func isNumber(b byte) bool {
+	return b >= '0' && b <= '9'
+}
+
+func validate(in configInput) error {
+	if in.configScope != "" {
+		maxLen := 64
+		configScopeLen := len(in.configScope)
+		if configScopeLen > maxLen {
+			return fmt.Errorf("config-scope must not exceed %d characters. Current length: %d\n", maxLen, configScopeLen)
+		}
+
+		if !isLetter(in.configScope[0]) {
+			return fmt.Errorf("config-scope must begin with a letter\n")
+		}
+
+		for i := 1; i < configScopeLen; i++ {
+			if !isLetter(in.configScope[i]) && !isNumber(in.configScope[i]) && in.configScope[i] != '-' {
+				return fmt.Errorf("config-scope must only contain letters, numbers, or '-'. Found '%c'.\n", in.configScope[i])
+			}
+		}
+	}
+
+	return nil
 }
 
 func generate(in configInput) ([]byte, error) {
--- a/main.go
+++ b/main.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"regexp"
 	"strconv"
 	"time"
 
@@ -126,10 +127,8 @@ func main() {
 		configScope:        *configScope,
 	}
 
-	validationErr := validate(input)
-
-	if validationErr != nil {
-		fmt.Fprintf(os.Stderr, "Error: %s\n", validationErr)
+	if err := validate(input); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %s\n", err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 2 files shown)


gnossen:
@ejona86 I guess you're the de facto reviewer for this repo? Please take a look.

ejona86:
The general code here looks fine. But is there any design document that actually describes the expected behavior here? It doesn't have to be a client-side design, but something talking about TD consuming these two strings would be helpful as we wonder "why did we do that" in the future.

main.go
```diff
@@ -190,14 +193,24 @@ func generate(in configInput) ([]byte, error) {
...
 		c.Node.Metadata[k] = v
 	}
 	if in.includeV3Features {
 		// xDS v2 implementation in TD expects the projectNumber and networkName in
 		// the metadata field while the v3 implementation expects these in the id
 		// field.
-		c.Node.Id = fmt.Sprintf("projects/%d/networks/%s/nodes/%s", in.gcpProjectNumber, in.vpcNetworkName, uuid.New().String())
+		var networkIdentifier string
+		if in.configScope != "" {
+			networkIdentifier = fmt.Sprintf("scope:%s", in.configScope)
```
ejona86, line 209: Should we restrict the config scope to only certain characters?

ejona86:
Based on another conversation, it sounded like this may be intended to be short-term. If so, that sounds fine and we don't need to be careful about it. But it'd be useful to call that out.

main.go
gnossen: Good point. Doing a little digging, it looks like scope has a maximum length of 64 and must conform to the regex `[a-zA-Z][-a-zA-Z0-9]*`. This validation is done in CLH, but it wouldn't hurt to add it here as well. Done.

gnossen:
Apparently, there isn't a design doc, but for posterity, this is [the implementation CL](cl/393194973).

PTALA.

main.go
```diff
@@ -122,7 +123,17 @@ func main() {
 		secretsDir:         *secretsDir,
 		metadataLabels:     nodeMetadata,
 		deploymentInfo:     deploymentInfo,
-	})
+		configScope:        *configScope,
+	}
+
+	validationErr := validate(input)
```
ejona86, line 129: Combine this assignment with the if, so that the variable goes out of scope? Err... actually, this is just a normal error, so leave it as-is but rename the variable to `err`.

main.go
```diff
@@ -165,6 +176,37 @@ type configInput struct {
...
+		configScopeLen := len(in.configScope)
+		if configScopeLen > maxLen {
+			return fmt.Errorf("config-scope must not exceed %d characters. Current length: %d\n", maxLen, configScopeLen)
+		}
+
+		if !isLetter(in.configScope[0]) {
+			return fmt.Errorf("config-scope must begin with a letter\n")
+		}
+
+		for i := 1; i < configScopeLen; i++ {
```
ejona86, line 202: `for _, c := range configScope {` ?

ejona86 (commented):
To be honest, I'd just expected a regex (`[a-zA-Z][a-zA-Z0-9-]{0,63}`) and we give an error "config-scope may only contain letters, numbers, and '-'. It must begin with a letter and must exceed 64 characters in length". But this works too.

easwars:
> To be honest, I'd just expected a regex (`[a-zA-Z][a-zA-Z0-9-]{0,63}`) and we give an error "config-scope may only contain letters, numbers, and '-'. It must begin with a letter and must exceed 64 characters in length". But this works too.

I would also prefer a regex. See: https://play.golang.org/p/_xO-VPmwFBo

gnossen:
Comments addressed. Please take another look. @easwars @ejona86

main.go
```diff
@@ -165,6 +175,17 @@ type configInput struct {
...
 	metadataLabels     map[string]string
 	deploymentInfo     map[string]string
+	configScope        string
+}
+
+func validate(in configInput) error {
+	re := regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]{0,63}$`)
+	if in.configScope != "" && !re.MatchString(in.configScope) {
+		return fmt.Errorf("config-scope may only contain letters, numbers, and '-'. " +
+			"It must begin with a letter and must exceed 64 characters in length")
```
ejona86, line 185: must _not_ exceed

ejona86 (approved):

main.go
gnossen: Indeed. What an odd restriction that would be.

ejona86:
@easwars, does it look good to you? (You are more well-versed in Go readability and such.)

main.go
```diff
@@ -165,6 +175,17 @@ type configInput struct {
 	secretsDir         string
 	metadataLabels     map[string]string
 	deploymentInfo     map[string]string
+	configScope        string
+}
+
+func validate(in configInput) error {
+	re := regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]{0,63}$`)
+	if in.configScope != "" && !re.MatchString(in.configScope) {
+		return fmt.Errorf("config-scope may only contain letters, numbers, and '-'. " +
```
easwars, line 184: Nit: It's totally acceptable to have insanely long lines in Go. Prefer not splitting lines up unless it improves readability. Thanks.

easwars:
> @easwars, does it look good to you? (You are more well-versed in Go readability and such.)

I thought this behavior of failing to hit the final reply button, and your comments sitting in pending state happens only on critique. Proved wrong today.

Just made some minor nit comments. Otherwise, the changes look good to me. Will approve once the nits are taken care of.

main.go
```diff
@@ -190,14 +211,24 @@ func generate(in configInput) ([]byte, error) {
...
+
 	for k, v := range in.metadataLabels {
 		c.Node.Metadata[k] = v
 	}
 	if in.includeV3Features {
 		// xDS v2 implementation in TD expects the projectNumber and networkName in
 		// the metadata field while the v3 implementation expects these in the id
 		// field.
-		c.Node.Id = fmt.Sprintf("projects/%d/networks/%s/nodes/%s", in.gcpProjectNumber, in.vpcNetworkName, uuid.New().String())
+		var networkIdentifier string
```
easwars, line 225: One more small nit. This is more idiomatic Go.

```golang
networkIdentifier := in.vpcNetworkName
if in.configScope != "" {
	networkIdentifier = fmt.Sprintf("scope:%s", in.configScope)
}
```

gnossen:
Thanks for the reviews! I think all comments have been resolved.

main_test.go
```diff
@@ -16,16 +16,78 @@ package main
...
+			},
+			wantError: "config-scope may only contain letters, numbers, and '-'. It must begin with a letter and must not exceed 64 characters in length",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			err := validate(test.input)
+			if test.wantError != fmt.Sprint(err) {
+				t.Fatalf("validate(%+v) returned output does not match expected:\nWant: \"%s\"\nGot: \"%s\"", test.input, test.wantError, fmt.Sprint(err))
```
easwars, line 85: It's always got before want.
go/go-style/decisions#got-before-want

main_test.go
```diff
@@ -16,16 +16,78 @@ package main
...
+				configScope:       "h*x8",
+			},
+			wantError: "config-scope may only contain letters, numbers, and '-'. It must begin with a letter and must not exceed 64 characters in length",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			err := validate(test.input)
+			if test.wantError != fmt.Sprint(err) {
```
easwars, line 84: I would use `err.Error()` instead of `fmt.Sprintf(err)`.

And the more common way to print an `err` here would be:
t.Fatalf("blah foo: %v", err)
https://tour.golang.org/methods/19

easwars (approved):
Sorry I failed to notice these in the previous iteration.

main_test.go
gnossen: Done.

PR #24, opened 2021-11-19
Drop xDS v2 support from bootstrap generator

gnossen:
App Net will no longer be supporting xDS V2, so we drop the v2-compatible method of supplying scope in the bootstrap.

@ejona86 @easwars

### DIFF
--- a/main.go
+++ b/main.go
@@ -216,10 +216,6 @@ func generate(in configInput) ([]byte, error) {
 		},
 	}
 
-	if in.configScope != "" {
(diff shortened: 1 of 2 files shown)


gnossen:
@ejona86 @easwars

ejona86 (approved):

easwars:
The PR title made me think that this is getting rid of xDS v2 support completely. But looks like it is only affecting one particular field. Could you please change the PR title.

easwars:
Oops ... it is already merged

PR #26, opened 2021-12-30
Rename --config-scope-exerimental to --config-mesh-experimental

gnossen:
This also includes a change from a "scope:" prefix in the node ID to a "mesh:" prefix instead.


gnossen:
@easwars

easwars:
We have had a release with this flag. 
@ejona86 : Is it OK to change the name now?

gnossen:
The name includes `-experimental` and the help text says "This flag is EXPERIMENTAL and may be changed or removed in a later release.". If this change isn't okay, then we need to change our messaging.

main.go
```diff
@@ -130,7 +130,7 @@ func main() {
 		secretsDir:         *secretsDir,
 		metadataLabels:     nodeMetadata,
 		deploymentInfo:     deploymentInfo,
-		configScope:        *configScope,
+		configMesh:        *configMesh,
```
ejona86, line 133: Please run gofmt.

ejona86 (approved):
Backward compatibility is no concern, excepting maybe anything we may do to make migrating our own tests easier (which is probably nothing, since we pin the version used in the tests). gofmt does need to be run though.

main.go
gnossen: Done.

easwars (approved):

PR #27, opened 2022-02-28
add basic support for federation

easwars:
Federation support in grpc is described [here](https://github.com/grpc/proposal/blob/master/A47-xds-federation.md)

This PR does the following:
- introduces a `--include-federation-experimental` flag to enable generation of configs to support federation. 
- adds an entry to the `authorities` map with the empty string as key, and an empty `xds_servers` field.
  - any resources using new-style names with without an authority will match this entry and will use the same configuration as that of the top-level `xds_servers` field.

In a follow-up PR, I will add support (via another command line flag) for the user to specify a JSON representation of the configuration for the `authorities` field. This will enable them to use custom authorities.


easwars:
Gentle ping ...

ejona86:
Spoke "offline." It isn't clear to me what this change is supposed to address ("what fails to work if we don't do this"). I'd have expected us to add a real authority for TD, if anything.

ejona86:
Closing for now. We can reopen when we know what we need.

PR #28, opened 2022-04-21
Replace Travis-CI with GitHub Actions

ejona86:
This tries to be 1:1 with no behavior differences except where it is
run.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,20 @@
+name: Testing
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/testing.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.14
+
+      - name: Run tests
+        run: go test -v ./...
```
easwars, line 20: Should we also add an action which runs `gofmt` or `go vet`?

ejona86: Maybe, but "This tries to be 1:1 with no behavior differences except where it is run."

First I'd like to bump the go version.

easwars (approved):

PR #29, opened 2022-05-09
Use Go 1.18 in CI

ejona86:
Go 1.18 includes VCS information within the binary so it is easy to
determine the git commit the binary was built with using `go version
-m`. Go considers the VCS modified if there are unknown files, so add
gitignore to avoid causing the build to be considered dirty.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.14
+          go-version: 1.18
 
       - name: Run tests
         run: go test -v ./...
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/td-grpc-bootstrap
+/td-grpc-bootstrap-*.tar.gz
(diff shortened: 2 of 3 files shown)


easwars (commented):
Should we update the Go version in the `go.mod` as well?

ejona86:
I considered it, but it is really unrelated. Such a change implies changes to the build/binary behavior, which is not what I was targeting here. I do think getting to `go 1.17` would be a good idea as just general practice because of the module loading improvements, although it doesn't actually help anything in this case. Also I think we probably want to allow the binary to be built with go 1.17, for example, which is what I have on my machine (via Debian).

easwars (approved):

easwars:
I'm surprised that we don't use any Go features introduced after Go1.13 :)

ejona86:
This broke the build, because `go build` invokes `git` which is not present in the image.

ejona86:
> I'm surprised that we don't use any Go features introduced after Go1.13 :)

I don't think it checks for library changes. Just _incompatible_ language changes (which haven't happened yet) and some go mod tweaks. So nothing says it actually works on 1.13 :)

PR #30, opened 2022-05-09
Use non-Alpine image on Cloud Build CI

ejona86:
This fixes a regression introduced in 9f045adb that broke the Cloud
Build CI:

```
Step #0: go: missing Git command. See https://golang.org/s/gogetcmd
Step #0: error obtaining VCS status: exec: "git": executable file not found in $PATH
Step #0: 	Use -buildvcs=false to disable VCS stamping.
```

In Go 1.18 the Go command will invoke the git command to include VCS
information inside the binary. We want that functionality, so we don't
want to pass `-buildvcs=false`. The non-Alpine container includes the
git command.

### DIFF
--- a/tools/cloudbuild-artifacts.yaml
+++ b/tools/cloudbuild-artifacts.yaml
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 steps:
-- name: golang:1.18-alpine
+- name: golang:1.18
   args: ['go', 'build', './...']
-- name: golang:1.18-alpine
+- name: golang:1.18
   args: ['go', 'test', './...']
 - name: alpine
   args: ['./tools/package.sh', '$COMMIT_SHA']


easwars:
Is there a way to install `git` through some GA step. The size difference between the alpine and non-alpine images is huge. Would be great if we can continue to use the alpine image.

ejona86:
It isn't that big of a difference vs normal Alpine. It is 100 vs 300 mb or something instead of the normal very small Alpine. Ideally the containers would be cached, so it is better to use a container that has what we need instead of install on every build. Also, Cloud Build is really not oriented toward installing things on demand.

easwars:
> It isn't that big of a difference vs normal Alpine. It is 100 vs 300 mb

I tried building some docker images:
using `scratch`: 6.8 MB
using `golang:1.18-alpine`: 335 MB
using `golang:1.18`: 970 MB

> Also, Cloud Build is really not oriented toward installing things on demand.

Ack

easwars (approved):

PR #31, opened 2022-08-08
Add experimental support for ignore_resource_deletion

ejona86:
https://github.com/grpc/proposal/blob/master/A53-xds-ignore-resource-deletion.md

### DIFF
--- a/main.go
+++ b/main.go
@@ -39,6 +39,7 @@ var (
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName        = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone          = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -39,6 +39,7 @@ var (
 		"the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName        = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone          = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
+	ignoreDeletion        = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is not currently the case. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
easwars, line 42: Can we rename this variable as `ignoreResourceDeletion`. We usually try to keep the flag name and the variable name in sync (except for the experimental suffix).

easwars (approved):

main.go
ejona86: I chose the shorted variable name to avoid the reformatting spam. Updated to longer name. PTAL

easwars (approved):

main.go
easwars: Thanks.

Issue #32, opened 2022-10-26
Changes to support xDS Federation


arvindbr8:
Per [A47 xDS Federation](https://github.com/grpc/proposal/blob/master/A47-xds-federation.md) design proposal two new fields are added to the gRPC xDS bootstrap config format, `authorities` and  `client_default_listener_resource_name_template`. 

Here are the new formats 

+ authorities

```
// A map of authority name to corresponding configuration.
//
// This is used in the following cases:
// - A gRPC client channel is created using an "xds:" URI that includes
//   an authority.
// - A gRPC client channel is created using an "xds:" URI with no
//   authority, but the "client_default_listener_resource_name_template"
//   field turns it into an "xdstp:" URI.
// - A gRPC server is created and the
//   "server_listener_resource_name_template" field is an "xdstp:" URI.
//
// In any of those cases, it is an error if the specified authority is
// not present in this map.
"authorities": {
  // Entries are keyed by authority name.
  // Note: If a new-style resource name has no authority, we will use
  // the empty string here as the key.
  "<authority_name>": {
    // A template for the name of the Listener resource to subscribe
    // to for a gRPC client channel.  Used only when the channel is
    // created using an "xds:" URI with this authority name.
    //
    // The token "%s", if present in this string, will be replaced
    // with percent-encoded service authority (i.e., the path part of the
    // target URI used to create the gRPC channel).  The replacement string
    // must include only characters allowed in a URI path as per RFC-3986
    // section 3.3 (which includes '/'), and all other characters must be
    // percent-encoded.
    //
    // Must start with "xdstp://<authority_name>/".  If it does not,
    // that is considered a bootstrap file parsing error.
    //
    // If not present in the bootstrap file, defaults to
    // "xdstp://<authority_name>/envoy.config.listener.v3.Listener/%s".
    "client_listener_resource_name_template": <string>,

    // Ordered list of xDS servers to contact for this authority.
    // Format is exactly the same as the top level "xds_servers" field.
    //
    // If the same server is listed in multiple authorities, the
    // entries will be de-duped (i.e., resources for both authorities
    // will be fetched on the same ADS stream).
    //
    // If not specified, the top-level server list is used.
    "xds_servers": [ ... ]
  }
}
```

+ client_default_listener_resource_name_template

```
// A template for the name of the Listener resource to subscribe to
// for a gRPC client channel.  Used only when the channel is created
// with an "xds:" URI with no authority.
//
[...]

PR #33, opened 2022-10-26
Changes to support xDS Federation

arvindbr8:
as per a47 xDS Federation [spec](https://github.com/grpc/proposal/blob/master/A47-xds-federation.md) we need to add the authorities field. 
For the authorities field, we want two entries:

+ Empty string: This will be used for all new style resource names which contain no authority
+ "trafficdirector.googleapis.com:443": This is for new style resource names which explicitly specify TD
For both these entries, we will leave the value empty or {}. That way, they will still both end up talking to TD with the server config specified by the top-level

Note that in `authorities` map authority name's value is empty. This is to force default to the `xds_server_uri` specified in the top level `xds_servers` field. 

Fixes _task 1_ in #32

### DIFF
--- a/main.go
+++ b/main.go
@@ -394,11 +394,6 @@ type creds struct {
 	Config interface{} `json:"config,omitempty"`
 }
 
-type authority struct {
-	ClientListenerResourceNameTemplate string   `json:"client_listener_resource_name_template,omitempty"`
-	XdsServers                         []server `json:"xds_servers,omitempty"`
-}
-
 type node struct {
 	Id           string                 `json:"id,omitempty"`
 	Cluster      string                 `json:"cluster,omitempty"`
--- a/main.go
+++ b/main.go
@@ -265,8 +265,9 @@ func generate(in configInput) ([]byte, error) {
 	}
 
 	if in.includeFederationSupport {
-		// Authorities with an empty server config will end up using the top-level server config.
-		// For more details, see: https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes.
+		// Authorities with an empty server config will end up using
+		// the top-level server config. For more details, see:
+		// https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes.
 		c.Authorities = map[string]struct{}{
 			tdURI: {},
 			"":    {},
--- a/main.go
+++ b/main.go
@@ -32,23 +32,25 @@ import (
 	"github.com/google/uuid"
 )
 
+const TdUri = "trafficdirector.googleapis.com:443"
+
 var (
-	xdsServerUri           = flag.String("xds-server-uri", "trafficdirector.googleapis.com:443", "override of server uri, for testing")
-	outputName             = flag.String("output", "-", "output file name")
-	gcpProjectNumber       = flag.Int64("gcp-project-number", 0, "the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName         = flag.String("vpc-network-name", "default", "VPC network name")
-	localityZone           = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
-	ignoreResourceDeletion = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is not currently the case. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	includeV3Features      = flag.Bool("include-v3-features-experimental", true, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	includePSMSecurity     = flag.Bool("include-psm-security-experimental", true, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	secretsDir             = flag.String("secrets-dir", "/var/run/secrets/workload-spiffe-credentials", "path to a directory containing TLS certificates and keys required for PSM security")
-	includeDeploymentInfo  = flag.Bool("include-deployment-info-experimental", false, "whether or not to generate config which contains deployment related information. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	gkeClusterName         = flag.String("gke-cluster-name-experimental", "", "GKE cluster name to use, instead of retrieving it from the metadata server. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	gkePodName             = flag.String("gke-pod-name-experimental", "", "GKE pod name to use, instead of reading it from $HOSTNAME or /etc/hostname file. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	gkeNamespace           = flag.String("gke-namespace-experimental", "", "GKE namespace to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -32,23 +32,25 @@ import (
...
+	gkePodName                  = flag.String("gke-pod-name-experimental", "", "GKE pod name to use, instead of reading it from $HOSTNAME or /etc/hostname file. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	gkeNamespace                = flag.String("gke-namespace-experimental", "", "GKE namespace to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	gceVM                       = flag.String("gce-vm-experimental", "", "GCE VM name to use, instead of reading it from the metadata server. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	configMesh                  = flag.String("config-mesh-experimental", "", "Dictates which Mesh resource to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	includeXDSFederationSupport = flag.Bool("include-federation-support-experimental", false, "whether or not to generate configs required for xDS Federation. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
easwars, line 53: s/includeXDSFederationSupport/includeFederationSupport/

Apart from the `experimental` suffix, flag variable names should match the flag names. So, skip the `XDS` in the flag variable name. We are anyways in deep xDS land here in the bootstrap generator, so we can safely ignore mentioning it.

main.go
```diff
@@ -185,6 +188,7 @@ type configInput struct {
 	metadataLabels         map[string]string
 	deploymentInfo         map[string]string
 	configMesh             string
+	includeXDSFederation   bool
```
easwars, line 191: Same here. Skip `XDS`.

main.go
```diff
@@ -362,6 +377,7 @@ func getFromMetadata(urlStr string) ([]byte, error) {
 
 type config struct {
 	XdsServers                         []server                             `json:"xds_servers,omitempty"`
+	Authorities                        map[string]authority                 `json:"authorities,omitempty"`
```
easwars: This won't compile because you have gotten rid of the `authority` struct now. Thinking about this more, I think it is better if we have the struct, but the struct can be an empty one, with a TODO saying, we will add fields to it when we make them customizable.

Because without the struct, this would have to be:
```golang
Authorities map[string]interface{}
```
Better to avoid the `interface{}` when we can.

main.go
```diff
@@ -260,6 +264,17 @@ func generate(in configInput) ([]byte, error) {
 		c.Node.Metadata["TRAFFIC_DIRECTOR_CLIENT_ENVIRONMENT"] = in.deploymentInfo
 	}
 
+	// Check for include XDS Federation flag to include Authorities in bootstrap config
```
easwars, line 267: Remove this comment? It is stating the obvious.

main.go
```diff
@@ -260,6 +264,17 @@ func generate(in configInput) ([]byte, error) {
...
+	if in.includeXDSFederation {
+		// both entries point to {} which is the top-level xds server config
```
easwars, line 269: Nit: Comments that are complete sentences should be capitalized and punctuated like standard English sentences. (As an exception, it is okay to begin a sentence with an uncapitalized identifier name if it is otherwise clear. Such cases are probably best done only at the beginning of a paragraph.) This comment applies to the two comment sentences below as well.
See: go/go-style/decisions#comment-sentences

Something like this would be nicer:
```golang
// Authorities with an empty server config will end up using the top-level server config. 
// For more details, see: https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes.

File: main.go [hunk 6]
```diff
@@ -32,23 +32,25 @@ import (
 	"github.com/google/uuid"
 )
 
+const TdUri = "trafficdirector.googleapis.com:443"
```
easwars, line 35: See: go/go-style/decisions#initialisms.

main_test.go
```diff
@@ -96,13 +96,14 @@ func TestGenerate(t *testing.T) {
...
-				metadataLabels:    map[string]string{"k1": "v1", "k2": "v2"},
-				includeV3Features: true,
+				xdsServerUri:         "example.com:443",
+				gcpProjectNumber:     123456789012345,
+				vpcNetworkName:       "thedefault",
+				ip:                   "10.9.8.7",
+				zone:                 "uscentral-5",
+				metadataLabels:       map[string]string{"k1": "v1", "k2": "v2"},
+				includeV3Features:    true,
+				includeXDSFederation: true,
```
easwars, line 106: Avoid changing this test. Instead add a new one like how each of the features has its own.

main.go
arvindbr8: sgtm

main.go
arvindbr8: got it. updating this

main_test.go
arvindbr8: sg

main.go
arvindbr8: sure. removing it

main.go
arvindbr8: I keep making this mistake. Updating it now

main_test.go
arvindbr8: my bad actually. adding one more v3 and v2 config cases

main.go
```diff
@@ -260,6 +264,15 @@ func generate(in configInput) ([]byte, error) {
 		c.Node.Metadata["TRAFFIC_DIRECTOR_CLIENT_ENVIRONMENT"] = in.deploymentInfo
 	}
 
+	if in.includeFederationSupport {
+		// Authorities with an empty server config will end up using the top-level server config.
+		// For more details, see: https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes.
```
easwars, line 269: Nit: Can you please ensure that comment lines are readable on narrow screens. See: go/go-style/decisions#comment-line-length.

main_test.go
```diff
@@ -402,6 +402,90 @@ func TestGenerate(t *testing.T) {
       "zone": "uscentral-5"
     }
   }
+}`,
+		},
+
```
easwars, line 407: Nit: nix newline

main_test.go
```diff
@@ -402,6 +402,90 @@ func TestGenerate(t *testing.T) {
...
+    "id": "52fdfc07-2182-454f-963f-5f0f9a621d72~10.9.8.7",
+    "cluster": "cluster",
+    "metadata": {
+      "TRAFFICDIRECTOR_GCP_PROJECT_NUMBER": "123456789012345",
+      "TRAFFICDIRECTOR_NETWORK_NAME": "thedefault"
+    },
+    "locality": {
+      "zone": "uscentral-5"
+    }
+  }
```
easwars, line 488: I don't think we need this case as none of the logic added as part of this PR depends on whether or not v3 features are enabled.

main_test.go
arvindbr8: 👍

main.go
arvindbr8: done 👍

main_test.go
arvindbr8: yup that makes sense to me

main.go
arvindbr8: done. updated

easwars (approved):

easwars:
@ejona86 : I have reviewed this PR and looks fine to me. Do you mind taking a quick look as well? Thanks.

ejona86 (approved):
The changes look fine, but this doesn't actually fix #32 entirely, correct, because we want the c2p stuff to be present?

easwars:
> The changes look fine, but this doesn't actually fix #32 entirely, correct, because we want the c2p stuff to be present?

Yes, this change only addresses basic federation. The c2p changes would follow after this.

arvindbr8:
Yes I can add c2p changes as a separate task in #32

PR #34, opened 2022-11-08
Add support to include c2p authority

arvindbr8:
As part of xds federation changes, we need to add support to the bootstrap generator to add c2p authority to the config. This change addresses that. 

### what's happenin?
1. add C2P TD authority option using a `include-directpath-authority-experimental` flag 
2. add additional logic to the bootstrap generator to query `http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s` and set the `TRAFFICDIRECTOR_DIRECTPATH_C2P_IPV6_CAPABLE` to true or false, in node metadata

Part of #32

### DIFF
--- a/main.go
+++ b/main.go
@@ -210,7 +210,7 @@ func validate(in configInput) error {
 
 func generate(in configInput) ([]byte, error) {
 	c := &config{
-		XdsServers: generateServerConfigsFromInputs([]string{in.xdsServerUri}, in),
+		XdsServers: generateServerConfigsFromInputs(in.xdsServerUri, in),
 		Node: &node{
 			Id:      uuid.New().String() + "~" + in.ip,
 			Cluster: "cluster", // unused by TD
@@ -270,7 +270,7 @@ func generate(in configInput) ([]byte, error) {
 		}
 		if in.includeDirectPathAuthority {
 			c.Authorities[c2pAuthority] = Authority{
-				XdsServers: generateServerConfigsFromInputs([]string{"dns:///directpath-pa.googleapis.com"}, in),
+				XdsServers: generateServerConfigsFromInputs("dns:///directpath-pa.googleapis.com", in),
 			}
 			if in.ipv6Capable {
 				c.Node.Metadata["TRAFFICDIRECTOR_DIRECTPATH_C2P_IPV6_CAPABLE"] = true
@@ -351,8 +351,8 @@ func getVMName() string {
 	return string(vm)
 }
 
-// isIPv6Capable. Retrieve IPv6 address (if any) from metadata server to check
-// for DirectPath IPv6 capability.
+// isIPv6Capable returns true if the VM is configured with an IPv6 address.
+// This will contact the metadata server to retrieve this information.
 func isIPv6Capable() bool {
 	_, err := getFromMetadata("http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s")
 	return err == nil
@@ -374,16 +374,16 @@ func getFromMetadata(urlStr string) ([]byte, error) {
 		},
 	}
 	resp, err := client.Do(req)
-	defer resp.Body.Close()
 	if err != nil {
 		return nil, fmt.Errorf("failed communicating with metadata server: %w", err)
 	}
 	body, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
 	if err != nil {
 		return nil, fmt.Errorf("failed reading from metadata server: %w", err)
 	}
-	if statusOk := resp.StatusCode >= 200 && resp.StatusCode < 300; !statusOk {
-		return nil, fmt.Errorf("failed reading from metadata server with Non-OK status code: %d", resp.StatusCode)
+	if code := resp.StatusCode; code < 200 || code > 299 {
+		return nil, fmt.Errorf("failed reading from metadata server with Non-OK status code: %d", code)
 	}
 	return body, nil
 }
@@ -402,25 +402,21 @@ type server struct {
 	ServerFeatures []string `json:"server_features,omitempty"`
 }
 
-func generateServerConfigsFromInputs(serverUris []string, in configInput) []server {
-	var ss []server
-	for _, su := range serverUris {
-		s := server{
-			ServerUri: su,
-			ChannelCreds: []creds{
-				{Type: "google_default"},
-			},
-		}
-		if in.includeV3Features {
-			s.ServerFeatures = append(s.ServerFeatures, "xds_v3")
-		}
-		if in.ignoreResourceDeletion {
-			s.ServerFeatures = append(s.ServerFeatures, "ignore_resource_deletion")
-		}
-		ss = append(ss, s)
+func generateServerConfigsFromInputs(serverUri string, in configInput) []server {
+	s := server{
+		ServerUri: serverUri,
+		ChannelCreds: []creds{
+			{Type: "google_default"},
+		},
+	}
+	if in.includeV3Features {
+		s.ServerFeatures = append(s.ServerFeatures, "xds_v3")
+	}
+	if in.ignoreResourceDeletion {
+		s.ServerFeatures = append(s.ServerFeatures, "ignore_resource_deletion")
 	}
 
-	return ss
+	return []server{s}
 }
 
 // Authority is the configuration corresponding to an authority name in the map.
--- a/main.go
+++ b/main.go
@@ -383,7 +383,7 @@ func getFromMetadata(urlStr string) ([]byte, error) {
 		return nil, fmt.Errorf("failed reading from metadata server: %w", err)
 	}
 	if code := resp.StatusCode; code < 200 || code > 299 {
-		return nil, fmt.Errorf("failed reading from metadata server with Non-OK status code: %d", code)
+		return nil, fmt.Errorf("metadata server returned status code %d for url %q", code, parsedUrl)
 	}
 	return body, nil
 }
--- a/main.go
+++ b/main.go
@@ -32,27 +32,29 @@ import (
 	"github.com/google/uuid"
 )
 
-const tdURI = "trafficdirector.googleapis.com:443"
-const c2pAuthority = "traffic-director-c2p.xds.googleapis.com"
+const (
+	tdURI        = "trafficdirector.googleapis.com:443"
+	c2pAuthority = "traffic-director-c2p.xds.googleapis.com"
+)
 
 var (
-	xdsServerUri             = flag.String("xds-server-uri", tdURI, "override of server uri, for testing")
-	outputName               = flag.String("output", "-", "output file name")
-	gcpProjectNumber         = flag.Int64("gcp-project-number", 0, "the gcp project number. If unknown, can be found via 'gcloud projects list'")
-	vpcNetworkName           = flag.String("vpc-network-name", "default", "VPC network name")
-	localityZone             = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
-	ignoreResourceDeletion   = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is not currently the case. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	includeV3Features        = flag.Bool("include-v3-features-experimental", true, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	includePSMSecurity       = flag.Bool("include-psm-security-experimental", true, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	secretsDir               = flag.String("secrets-dir", "/var/run/secrets/workload-spiffe-credentials", "path to a directory containing TLS certificates and keys required for PSM security")
-	includeDeploymentInfo    = flag.Bool("include-deployment-info-experimental", false, "whether or not to generate config which contains deployment related information. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	gkeClusterName           = flag.String("gke-cluster-name-experimental", "", "GKE cluster name to use, instead of retrieving it from the metadata server. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	gkePodName               = flag.String("gke-pod-name-experimental", "", "GKE pod name to use, instead of reading it from $HOSTNAME or /etc/hostname file. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	gkeNamespace             = flag.String("gke-namespace-experimental", "", "GKE namespace to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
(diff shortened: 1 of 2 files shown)


arvindbr8:
@easwars please take a look

main.go
```diff
@@ -388,6 +395,10 @@ type server struct {
 	ServerFeatures []string `json:"server_features,omitempty"`
 }
 
+// Authority
```
easwars, line 398: Please add a complete comment.

main.go
```diff
@@ -268,10 +272,13 @@ func generate(in configInput) ([]byte, error) {
 		// Authorities with an empty server config will end up using
 		// the top-level server config. For more details, see:
 		// https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes.
-		c.Authorities = map[string]struct{}{
+		c.Authorities = map[string]Authority{
 			tdURI: {},
 			"":    {},
 		}
+		if in.includeC2PAuthority {
+			c.Authorities[c2pAuthority] = Authority{}
```
easwars, line 280: We don't want this to be empty, right. If it is empty, this authority will also use the top-level server config which will send the xDS requests to `trafficdirector.googleapis.com:443`. We want requests matching this authority to go to `dns:///directpath-pa.googleapis.com` instead.

easwars (changes requested):

main.go
arvindbr8: it didnt let me do the TODO without the `authority` in there. i'll add something in

main.go
arvindbr8: oh right and use DP. I think I was just talking to @apolycn about it now

main.go
```diff
@@ -33,6 +33,7 @@ import (
 )
 
 const tdURI = "trafficdirector.googleapis.com:443"
+const c2pAuthority = "traffic-director-c2p.xds.googleapis.com"
```
easwars, line 36: Nit: create a const block?

```golang
const (
    tdURI = "trafficdirector.googleapis.com:443"
    c2pAuthority = "traffic-director-c2p.xds.googleapis.com"
)
```

main.go
```diff
@@ -51,6 +52,7 @@ var (
 	gceVM                    = flag.String("gce-vm-experimental", "", "GCE VM name to use, instead of reading it from the metadata server. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	configMesh               = flag.String("config-mesh-experimental", "", "Dictates which Mesh resource to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	includeFederationSupport = flag.Bool("include-federation-support-experimental", false, "whether or not to generate configs required for xDS Federation. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	includeC2PAuthority      = flag.Bool("include-c2p-authority-experimental", false, "whether or not to include c2p TD authority for xDS Federation. Ignored if not used with include-federation-support-experimental flag. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
easwars, line 55: `c2p` might not be very descriptive for the users. Maybe `include-directpath-authority-experimental`. Wdyt @ejona86 ?

main.go
```diff
@@ -388,6 +404,12 @@ type server struct {
 	ServerFeatures []string `json:"server_features,omitempty"`
 }
 
+// Authority is the corresponding configuration to an authority name in the map.
```
easwars, line 407: Maybe:

```golang
// Authority is the configuration corresponding to an authority name in the map.
//
// For more details, see:
// https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes
```

main.go
```diff
@@ -268,10 +272,22 @@ func generate(in configInput) ([]byte, error) {
...
 			tdURI: {},
 			"":    {},
 		}
+		if in.includeC2PAuthority {
+			c.Authorities[c2pAuthority] = Authority{
+				XdsServers: []server{
+					{
+						ServerUri: "dns:///directpath-pa.googleapis.com",
+						ChannelCreds: []creds{
+							{Type: "google_default"},
```
easwars, line 285: We need server features with `v3` support for c2p TD as well.

main.go
arvindbr8: done :)

main.go
arvindbr8: yup, that makes sense

main.go
arvindbr8: sure, updated. So do we also want to check for `ignore_resource_deletion` too?

arvindbr8:
I'm also updating this to include changes to support adding node metadata indicating IPv6 capabilities for DirectPath [b/258291928](https://b.corp.google.com/issues/258291928)

main.go
easwars: Hmm ... maybe it might be a good idea to pull this out into a helper function. That way, we can be sure that we use the same logic to generate the server config for both TDs.

main.go
```diff
@@ -203,12 +211,7 @@ func validate(in configInput) error {
 func generate(in configInput) ([]byte, error) {
 	c := &config{
 		XdsServers: []server{
-			{
-				ServerUri: in.xdsServerUri,
-				ChannelCreds: []creds{
-					{Type: "google_default"},
-				},
-			},
+			generateServerConfigFromInput(in.xdsServerUri, in),
```
easwars, line 214: I'm wondering if we could make `generateServerConfigFromInput` return a `[]server` with a single entry instead of the caller doing that part. If we ever have to end up supporting more than one entry in the server config, we can deal with that then. What do you think?

main.go
```diff
@@ -347,6 +355,15 @@ func getVMName() string {
 	return string(vm)
 }
 
+// isIPv6Capable. Retrieve IPv6 address (if any) from metadata server to check
+// for DirectPath IPv6 capability.
+func isIPv6Capable() bool {
+	if _, err := getFromMetadata("http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s"); err == nil {
+		return true
+	}
+	return false
```
easwars, line 364: Optional:

```golang
    _, err := getFromMetadata("http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s"); 
    return err == nil
```

main_test.go
```diff
@@ -544,6 +605,24 @@ func TestGetVMName(t *testing.T) {
 	}
 }
 
+func TestCheckIPv6Capable(t *testing.T) {
+	server := httptest.NewServer(nil)
+	defer server.Close()
+	overrideHTTP(server)
+	want := true
```
easwars, line 612: Nit: move `want` to be declared right above where it is used.

main_test.go
```diff
@@ -544,6 +605,24 @@ func TestGetVMName(t *testing.T) {
...
+	http.HandleFunc("http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s",
+		func(w http.ResponseWriter, r *http.Request) {
+			if r.Header.Get("Metadata-Flavor") != "Google" {
+				http.Error(w, "Missing Metadata-Flavor", 403)
+				return
+			}
+			w.Write([]byte("6970:7636:2061:6464:7265:7373:2062:6162"))
+		})
+	if got := isIPv6Capable(); got != want {
```
easwars, line 621: Should we have a test case where `isIPv6Capable()` returns false?

I see that other tests don't do that. I don't remember why. But, if we *can* test both cases here, that would be good I think.

easwars (commented):
I think changes are mostly looking good. Some very minor comments this time around.

main.go
arvindbr8: looks better.. im updating this

main_test.go
arvindbr8: makes sense.. updating this

main.go
```diff
@@ -388,6 +399,36 @@ type server struct {
...
+	return ss
+}
+
+// Authority is the configuration corresponding to an authority name in the map.
+//
+// For more details, see:
+// https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes
+type Authority struct {
+	XdsServers                         []server `json:"xds_servers,omitempty"`
+	ClientListenerResourceNameTemplate string   `json:"client_listener_resource_name_template,omitempty"`
```
easwars, line 429: Do we need this field at this point, given that we are not using it?

main_test.go
```diff
@@ -544,6 +605,49 @@ func TestGetVMName(t *testing.T) {
...
+			},
+			wantOutput: false,
+		},
+	}
+	server := httptest.NewServer(nil)
+	defer server.Close()
+	overrideHTTP(server)
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			uuid.SetRand(rand.New(rand.NewSource(1)))
```
easwars, line 639: What is this required?

main_test.go
```diff
@@ -544,6 +605,49 @@ func TestGetVMName(t *testing.T) {
...
+			http.DefaultServeMux = new(http.ServeMux)
+			http.HandleFunc("http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s", test.httpHandler)
+			want := true
```
easwars, line 642: Shouldn't you be using `test.wantOutput` instead?

main_test.go
```diff
@@ -544,6 +605,49 @@ func TestGetVMName(t *testing.T) {
...
 }
 
+func TestCheckIPv6Capable(t *testing.T) {
+	tests := []struct {
+		desc        string
+		httpHandler func(http.ResponseWriter, *http.Request)
+		wantOutput  bool
+	}{
+		{
+			desc: "case with check on ipv6 enabled host",
```
easwars, line 615: See: go/go-style/decisions#subtest-names

I'm thinking `v6 enabled` and `v6 not enabled`.

main_test.go
arvindbr8: yup thanks for catching that. when i fixed that it seems like there was a bug.. 
I was checking for err returned from `getFromMetadata`, but `getFromMetadata` does not return err for non-OK responses, and we return only the body from `getFromMetadata` -- so there is no way to check for status code from the caller. 
i'm adding addition logic in `getFromMetadata` to check for non-OK status codes. Lemeno what you think..

arvindbr8 (commented):
rev'ing the PR addressing comments. @easwars pls take a look when you can

main.go
```diff
@@ -347,6 +351,13 @@ func getVMName() string {
 	return string(vm)
 }
 
+// isIPv6Capable. Retrieve IPv6 address (if any) from metadata server to check
+// for DirectPath IPv6 capability.
```
easwars, line 355: How about:

```golang
// isIPv6Capable returns true if the VM is configured with an IPv6 address. 
// It contacts the metadata server to get this information.
```

main.go
```diff
@@ -363,20 +374,23 @@ func getFromMetadata(urlStr string) ([]byte, error) {
 		},
 	}
 	resp, err := client.Do(req)
+	defer resp.Body.Close()
```
easwars, line 377: Why is this change required?

According to the `http` package documentation for `Do()`:
```
On error, any Response can be ignored. A non-nil Response with a non-nil error only occurs when
CheckRedirect fails, and even then the returned Response.Body is already closed.
```

main.go
```diff
@@ -363,20 +374,23 @@ func getFromMetadata(urlStr string) ([]byte, error) {
...
 		return nil, fmt.Errorf("failed communicating with metadata server: %w", err)
 	}
 	body, err := ioutil.ReadAll(resp.Body)
-	resp.Body.Close()
 	if err != nil {
 		return nil, fmt.Errorf("failed reading from metadata server: %w", err)
 	}
+	if statusOk := resp.StatusCode >= 200 && resp.StatusCode < 300; !statusOk {
+		return nil, fmt.Errorf("failed reading from metadata server with Non-OK status code: %d", resp.StatusCode)
+	}
```
easwars, line 387: Maybe something like:

```golang
if code := resp.StatusCode; code < 200 || code > 299 {
	return nil, fmt.Errorf("metadata server returned status code %d for url %q", code, url)
}
```

I'm OK with the conditional being `!(code >= 200 && code < 300)` as well. But I feel the above one is simpler. Up to you.

main.go
```diff
@@ -388,6 +402,35 @@ type server struct {
 	ServerFeatures []string `json:"server_features,omitempty"`
 }
 
+func generateServerConfigsFromInputs(serverUris []string, in configInput) []server {
```
easwars, line 405: I understand this is future proof in the sense that it will handle the case where we want to support multiple server configs. But we dont have that need today and we don't foresee it in the near future. So, I would recommend changing this function to be `func generateServerConfigsFromInputs(serverUri string, in configInput) []server { ...}`.

See: https://engdoc.corp.google.com/eng/doc/tott/episodes/438.md?cl=head

main_test.go
```diff
@@ -544,6 +605,48 @@ func TestGetVMName(t *testing.T) {
...
+			},
+			wantOutput: false,
+		},
+	}
+	server := httptest.NewServer(nil)
+	defer server.Close()
+	overrideHTTP(server)
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			http.DefaultServeMux = new(http.ServeMux)
```
easwars, line 639: Do we still need this? It doesn't feel right to me to overwrite the package level variable `DefaultServeMux`.

If we really want to do something like this, it might involve a few more steps I think. The `httptest` package does not support using a serveMux other than the default one. So, we might have to create our own `net.Listener`, and pass it to [Serve](https://pkg.go.dev/net/http#Serve), along with a `Handler`. Would be nice if we can avoid all of that.

main.go
arvindbr8: :+1: makes sense.. Im updating this as well

main_test.go
arvindbr8: I had to do that because it throws an error since im trying to register a different handlerfunc for the same endpoint..

but this works

```
mux := new(http.ServeMux)
mux.HandleFunc("metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s", test.httpHandler)
server := httptest.NewServer(mux)
defer server.Close()
overrideHTTP(server)
if got := isIPv6Capable(); got != test.wantOutput {
 t.Fatalf("isIPv6Capable() = %t, want: %t", got, test.wantOutput)
}
```

lemeno if this is okayy

easwars (approved):

easwars:
@ejona86 : I think the changes look good. I'm assigning you for a second set of eyes. Thank you.

main.go
```diff
@@ -371,12 +382,15 @@ func getFromMetadata(urlStr string) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed reading from metadata server: %w", err)
 	}
+	if code := resp.StatusCode; code < 200 || code > 299 {
+		return nil, fmt.Errorf("failed reading from metadata server with Non-OK status code: %d", code)
```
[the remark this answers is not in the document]
arvindbr8: yup that makes sense.. I'll update that

main.go
ejona86: meh. Users should never need to set this. It'll just become the default and then go away.

ejona86 (approved):

PR #35, opened 2023-03-16
Add flag to use xdstp name for client listener resource template name

arvindbr8:
We added the `client_default_listener_resource_name_template` field to the bootstrap config as per  [A47](https://github.com/grpc/proposal/blob/master/A47-xds-federation.md). 

We would like to have the option for the bootstrap generator to add field to the authorities with the xdstp style name for listener resources with TD c2p authority name (when used in conjunction with `includeFederationSupport` and `includeDirectPathAuthority` flags. 

This PR adds a flag to support this. 

##### Other fix:
+ `trafficdirector.googleapis.com:443` seems wrong and should be `traffic-director-global.xds.googleapis.com` instead
  + When we do a lookup for which server to request for a CDS resource, it seems we wouldn't find a control plane

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,13 +33,12 @@ import (
 )
 
 const (
-	tdURI        = "trafficdirector.googleapis.com:443"
 	tdAuthority  = "traffic-director-global.xds.googleapis.com"
 	c2pAuthority = "traffic-director-c2p.xds.googleapis.com"
 )
 
 var (
-	xdsServerUri               = flag.String("xds-server-uri", tdURI, "override of server uri, for testing")
+	xdsServerUri               = flag.String("xds-server-uri", tdAuthority, "override of server uri, for testing")
 	outputName                 = flag.String("output", "-", "output file name")
 	gcpProjectNumber           = flag.Int64("gcp-project-number", 0, "the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName             = flag.String("vpc-network-name", "default", "VPC network name")
@@ -270,13 +269,13 @@ func generate(in configInput) ([]byte, error) {
 		// the top-level server config. For more details, see:
 		// https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes.
 		c.Authorities = map[string]Authority{
-			tdURI: {},
-			"":    {},
+			tdAuthority: {},
+			"":          {},
 		}
 		if in.includeXDSTPNameInLDS {
-			if a, ok := c.Authorities[tdURI]; ok {
+			if a, ok := c.Authorities[tdAuthority]; ok {
 				a.ClientListenerResourceNameTemplate = fmt.Sprintf("xdstp://%s/envoy.config.listener.v3.Listener/%%s", tdAuthority)
-				c.Authorities[tdURI] = a
+				c.Authorities[tdAuthority] = a
 			}
 		}
 		if in.includeDirectPathAuthority {
--- a/main.go
+++ b/main.go
@@ -55,6 +55,8 @@ var (
 	configMesh                 = flag.String("config-mesh-experimental", "", "Dictates which Mesh resource to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	includeFederationSupport   = flag.Bool("include-federation-support-experimental", false, "whether or not to generate configs required for xDS Federation. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	includeDirectPathAuthority = flag.Bool("include-directpath-authority-experimental", false, "whether or not to include DirectPath TD authority for xDS Federation. Ignored if not used with include-federation-support-experimental flag. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	// TODO: default to true when TD supports xdstp style names.
+	includeXDSTPNameInLDS = flag.Bool("include-xdstp-name-in-lds-experimental", false, "whether or not to use XDSTP name for . Ignored if not used with both include-federation-support-experimental and include-directpath-authority-experimental flag. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 )
 
 func main() {
@@ -197,6 +199,7 @@ type configInput struct {
 	includeFederationSupport   bool
 	includeDirectPathAuthority bool
 	ipv6Capable                bool
+	includeXDSTPNameInLDS      bool
 }
 
 func validate(in configInput) error {
@@ -269,9 +272,14 @@ func generate(in configInput) ([]byte, error) {
 			"":    {},
 		}
 		if in.includeDirectPathAuthority {
-			c.Authorities[c2pAuthority] = Authority{
+			authority := Authority{
 				XdsServers: generateServerConfigsFromInputs("dns:///directpath-pa.googleapis.com", in),
 			}
+			if in.includeXDSTPNameInLDS {
+				template := fmt.Sprintf("xdstp://%s/envoy.config.listener.v3.Listener/%%s", c2pAuthority)
+				authority.ClientListenerResourceNameTemplate = template
+			}
+			c.Authorities[c2pAuthority] = authority
 			if in.ipv6Capable {
 				c.Node.Metadata["TRAFFICDIRECTOR_DIRECTPATH_C2P_IPV6_CAPABLE"] = true
 			}
@@ -424,7 +432,8 @@ func generateServerConfigsFromInputs(serverUri string, in configInput) []server 
 // For more details, see:
 // https://github.com/grpc/proposal/blob/master/A47-xds-federation.md#bootstrap-config-changes
 type Authority struct {
-	XdsServers []server `json:"xds_servers,omitempty"`
+	XdsServers                         []server `json:"xds_servers,omitempty"`
+	ClientListenerResourceNameTemplate string   `json:"client_listener_resource_name_template,omitempty"`
 }
 
 type creds struct {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -55,6 +55,8 @@ var (
 	configMesh                 = flag.String("config-mesh-experimental", "", "Dictates which Mesh resource to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	includeFederationSupport   = flag.Bool("include-federation-support-experimental", false, "whether or not to generate configs required for xDS Federation. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	includeDirectPathAuthority = flag.Bool("include-directpath-authority-experimental", false, "whether or not to include DirectPath TD authority for xDS Federation. Ignored if not used with include-federation-support-experimental flag. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	// TODO: default to true when TD supports xdstp style names.
+	includeXDSTPNameInLDS = flag.Bool("include-xdstp-name-in-lds-experimental", false, "whether or not to use XDSTP name for . Ignored if not used with both include-federation-support-experimental and include-directpath-authority-experimental flag. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
easwars, line 59: `whether or not to use XDSTP name for .`

This sentence seems to be incomplete.

main.go
```diff
@@ -269,9 +272,14 @@ func generate(in configInput) ([]byte, error) {
 			"":    {},
 		}
 		if in.includeDirectPathAuthority {
-			c.Authorities[c2pAuthority] = Authority{
+			authority := Authority{
 				XdsServers: generateServerConfigsFromInputs("dns:///directpath-pa.googleapis.com", in),
 			}
+			if in.includeXDSTPNameInLDS {
+				template := fmt.Sprintf("xdstp://%s/envoy.config.listener.v3.Listener/%%s", c2pAuthority)
+				authority.ClientListenerResourceNameTemplate = template
```
easwars, line 280: Nit: Should be fine to inline this and get rid of the `template` variable.

main.go
arvindbr8: whoops.. my bad. I was in the processing of adding a sentence there and I think I just went along with the rest of the message. I've updated this.

main.go
arvindbr8: done. I was trying to keep the line width to 80.. but i guess there are multiple places in this file where that doesn't apply

main.go
```diff
@@ -269,9 +273,13 @@ func generate(in configInput) ([]byte, error) {
 			"":    {},
 		}
 		if in.includeDirectPathAuthority {
-			c.Authorities[c2pAuthority] = Authority{
+			authority := Authority{
 				XdsServers: generateServerConfigsFromInputs("dns:///directpath-pa.googleapis.com", in),
 			}
+			if in.includeXDSTPNameInLDS {
+				authority.ClientListenerResourceNameTemplate = fmt.Sprintf("xdstp://%s/envoy.config.listener.v3.Listener/%%s", c2pAuthority)
```
ejona86, line 280: I'm surprised to see this for c2p. I had expected the flag to be for plain TD and that c2p would be hard-coded to use xdstp. Looking at `GoogleCloudToProdNameResolver` in Java though, it seems we aren't setting `client_listener_resource_name_template` today. We should discuss what is going on, because I'm missing some context.

main.go
easwars: There is no need to keep code lines to 80 cols. That guideline is only for comment lines. 

See: https://g3doc.corp.google.com/go/g3doc/style/guide.md?cl=head#line-length and go/go-style/decisions#comment-line-length.

main.go
easwars: The c2p resolver in Go doesn't seem to be doing that either.

arvindbr8: I think you are right Eric, the flag should be for TD. But the authority name that I got from the TD team initially was the c2p authority. But I got it confirmed that  we should be using `traffic-director-global.xds.googleapis.com` for plain TD and for c2p I think we should be overriding this in the client. Updating

ejona86: It seems you are modifying the wrong server configuration. (Or you need to modify both, but with proper authorities)

arvindbr8: Sorry. I should have taken a second look. Seems to be correct now. 


PS: also realized i was pushing from my master. err...

main_test.go
```diff
@@ -506,6 +506,69 @@ func TestGenerate(t *testing.T) {
...
+              "type": "google_default"
+            }
+          ],
+          "server_features": [
+            "xds_v3"
+          ]
+        }
+      ]
+    },
+    "trafficdirector.googleapis.com:443": {}
```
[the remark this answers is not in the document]
arvindbr8: Done. I discussed this offline with jiechen@ and easwar@. This revision should be correct.

easwars (approved):
The changes look good to me. But you might have to rebase after #36. I believe your test might need some minor tweaking based on that.

main_test.go
```diff
@@ -506,6 +506,70 @@ func TestGenerate(t *testing.T) {
...
+            }
+          ],
+          "server_features": [
+            "xds_v3"
+          ]
+        }
+      ]
+    },
+    "trafficdirector.googleapis.com:443": {
+      "client_listener_resource_name_template": "xdstp://traffic-director-global.xds.googleapis.com/envoy.config.listener.v3.Listener/%s"
```
ejona86, line 557: Oh boy. This is going to cause confusion. It seems trafficdirector.googleapis.com is wrong, and it should be traffic-director-global.xds.googleapis.com instead. When we do a lookup for which server to request for a CDS resource, it seems we wouldn't find a control plane.

arvindbr8: I thought I got this verified by the TD team. I'll fix that as part of this PR itself.

ejona86: It is more of a grpc problem than a TD problem. It is completely within the scope of grpc.

main_test.go
```diff
@@ -463,7 +463,70 @@ func TestGenerate(t *testing.T) {
...
+				includeV3Features:          true,
+				includeFederationSupport:   true,
+				includeDirectPathAuthority: true,
+				ipv6Capable:                true,
+				includeXDSTPNameInLDS:      true,
+			},
+			wantOutput: `{
+  "xds_servers": [
+    {
+      "server_uri": "traffic-director-global.xds.googleapis.com",
```
ejona86, line 529: This is changing the domain name to use to contact TD. Is that correct?

You are changing this domain name even when the new experimental flag is _not_ being used, so this seems very dangerous.

arvindbr8:
Sorry about the delay here. Other tasks got prioritized. Please take a look now @ejona86

main_test.go
```diff
@@ -507,6 +507,71 @@ func TestGenerate(t *testing.T) {
...
+            }
+          ],
+          "server_features": [
+            "xds_v3"
+          ]
+        }
+      ],
+      "client_listener_resource_name_template": "xdstp://traffic-director-c2p.xds.googleapis.com/envoy.config.listener.v3.Listener/%s"
+    },
+    "trafficdirector.googleapis.com:443": {
```
ejona86, line 558: From my "Oh boy" comment, it seems that this needs to be traffic-director-global.xds.googleapis.com.

arvindbr8: I've got it updated now. Sorry for the back and forth.

main.go
```diff
@@ -265,9 +270,17 @@ func generate(in configInput) ([]byte, error) {
...
-			"":    {},
+			"": {},
 		}
+
+		if in.includeXDSTPNameInLDS {
+			c.Authorities[tdAuthority] = Authority{
+				ClientListenerResourceNameTemplate: fmt.Sprintf("xdstp://%s/envoy.config.listener.v3.Listener/%%s", tdAuthority),
+			}
+		} else {
+			c.Authorities[tdURI] = Authority{}
```
ejona86, line 281: We were saying we think this was a bug. Instead of s/tdURI/tdAuthority/ let's just remove this line entirely? Then the new flag being added in this PR adds the authority entry for tdAuthority.

ejona86 (approved):
I do want @easwars to take a look, since it has changed in an important way.

main.go
arvindbr8: Done. 👍

main.go
```diff
@@ -55,6 +56,8 @@ var (
 	configMesh                 = flag.String("config-mesh-experimental", "", "Dictates which Mesh resource to use. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	includeFederationSupport   = flag.Bool("include-federation-support-experimental", false, "whether or not to generate configs required for xDS Federation. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	includeDirectPathAuthority = flag.Bool("include-directpath-authority-experimental", false, "whether or not to include DirectPath TD authority for xDS Federation. Ignored if not used with include-federation-support-experimental flag. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	// TODO: default to true when TD supports xdstp style names.
+	includeXDSTPNameInLDS = flag.Bool("include-xdstp-name-in-lds-experimental", false, "whether or not to use XDSTP style name for listener resource name template. Ignored if not used with both include-federation-support-experimental flag. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
easwars, line 60: Nit: s/Ignored if not used with both include-federation-support-experimental flag/Ignored if not used with --include-federation-support-experimental flag/

easwars (approved):
LGTM.

main.go
arvindbr8: Done

PR #37, opened 2023-05-02
ignore_resource_deletion server feature is always set in directpath use-cases

arvindbr8:
Now in DIrectpath (C2P) usecases, `ignore_resource_deletion` server feature can always be set.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,20 +32,13 @@ import (
 	"github.com/google/uuid"
 )
 
-const (
-	tdURI        = "trafficdirector.googleapis.com:443"
-	c2pTdURI     = "dns:///directpath-pa.googleapis.com"
-	tdAuthority  = "traffic-director-global.xds.googleapis.com"
-	c2pAuthority = "traffic-director-c2p.xds.googleapis.com"
-)
-
 var (
-	xdsServerUri               = flag.String("xds-server-uri", tdURI, "override of server uri, for testing")
+	xdsServerUri               = flag.String("xds-server-uri", "trafficdirector.googleapis.com:443", "override of server uri, for testing")
 	outputName                 = flag.String("output", "-", "output file name")
 	gcpProjectNumber           = flag.Int64("gcp-project-number", 0, "the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName             = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone               = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
-	ignoreResourceDeletion     = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is only true for Google C2P (DirectPath) use-cases. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -44,7 +45,7 @@ var (
 	gcpProjectNumber           = flag.Int64("gcp-project-number", 0, "the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName             = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone               = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
-	ignoreResourceDeletion     = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is not currently the case. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	ignoreResourceDeletion     = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is only true for Google C2P (DirectPath) use-cases. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
ejona86, line 48: I think the previous language was okay, as DirectPath != Traffic Director from this perspective.

main.go
```diff
@@ -34,6 +34,7 @@ import (
 
 const (
 	tdURI        = "trafficdirector.googleapis.com:443"
+	c2pTdURI     = "dns:///directpath-pa.googleapis.com"
```
ejona86, line 37: `c2pURI` for consistency with `c2pAuthority`?

main.go
```diff
@@ -281,7 +282,9 @@ func generate(in configInput) ([]byte, error) {
 
 		if in.includeDirectPathAuthority {
 			c.Authorities[c2pAuthority] = Authority{
-				XdsServers:                         generateServerConfigsFromInputs("dns:///directpath-pa.googleapis.com", in),
+				// In the case of DirectPath, it is safe to assume that the operator is notified of missing resources.
+				// In other words, "ignore_resource_deletion" server_features is always set.
+				XdsServers:                         generateServerConfig(c2pTdURI, in.includeV3Features, true),
```
ejona86, line 287: @easwars, how would you feel about avoiding `generateServerConfigsFromInputs()` here and just hard-coding the config. I feel like using the function serves no purpose for DP. v3 should be hard-coded for it, and no conditionals at all. I feel like it got this way by trying too hard to reuse code.

main.go
arvindbr8: Sure. This only needs to change when we make the default `true`

main.go
arvindbr8: I agree. It seems like we have to go back and forth. and the helper is only used in two places. but for different reasons.
Done.

main.go
arvindbr8: I was talking to Easwar about something similar (in the context of another PR). This generator script is actually kinda long. Making URIs string constants are actually a bit annoying since it forces us to scroll back and forth for just this one value. And most of these consts only have a single usage. Seems like you've had a similar discussion with Easwar about this. 

I've actually nixed the consts and used the strings as is. @ejona86 lmk what you think about it?

arvindbr8 (commented):
Revised. @ejona86 PTAL :) 

PS: (I dont know if can promise you "enough" revisions like the last time)

ejona86 (approved):
I'm a very strong believer in inlining constants. I find the "no literals in code" coding view to be harmful. So this looks great.

easwars (approved):

PR #39, opened 2023-09-12
Fix LDS resource name format for TD authority

arvindbr8:
The LDS resource name format was incorrect. The correct format as per the xDS TD documentation is 

```
xdstp://<authority>/envoy.config.listener.v3.Listener/<project_number>/<(network)|(mesh:mesh_name)>/id
```

This change fixes the format.

### DIFF
--- a/main.go
+++ b/main.go
@@ -280,7 +280,7 @@ func generate(in configInput) ([]byte, error) {
 		if in.includeXDSTPNameInLDS {
 			tdAuthority := "traffic-director-global.xds.googleapis.com"
 			c.Authorities[tdAuthority] = Authority{
-				// Listener Resource Name format for p2c config looks like:
+				// Listener Resource Name format for normal TD usecases looks like:
 				// xdstp://<authority>/envoy.config.listener.v3.Listener/<project_number>/<(network)|(mesh:mesh_name)>/id
 				ClientListenerResourceNameTemplate: fmt.Sprintf("xdstp://%s/envoy.config.listener.v3.Listener/%d/%s/%%s", tdAuthority, in.gcpProjectNumber, networkIdentifier),
 			}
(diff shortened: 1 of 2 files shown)


easwars:
The PR description talks about `scope`. But there is no mention of it in the code (and in the resource name format mentioned in the comment in the code). If `scope` is something for which we will add support in the near future, can we keep the PR description and the code consistent (one way or the other)?

main.go
```diff
@@ -280,7 +280,9 @@ func generate(in configInput) ([]byte, error) {
 		if in.includeXDSTPNameInLDS {
 			tdAuthority := "traffic-director-global.xds.googleapis.com"
 			c.Authorities[tdAuthority] = Authority{
-				ClientListenerResourceNameTemplate: fmt.Sprintf("xdstp://%s/envoy.config.listener.v3.Listener/%%s", tdAuthority),
+				// Listener Resource Name format for p2c config looks like:
```
ejona86, line 283: Umm...
1. This entire generator isn't used for P2C
2. This is the normal TD case, no directpath involved

arvindbr8: According to the IYKYK/td-xds-federation document, `traffic-director-global.xds.googleapis.com` authority was in the entry that said P2C config. I was trying to stick to that name to find backref. 

So P2C means DirectPath but in the other direction?

arvindbr8:
> The PR description talks about scope. But there is no mention of it in the code (and in the resource name format mentioned in the comment in the code). If scope is something for which we will add support in the near future, can we keep the PR description and the code consistent (one way or the other)?

IIUC, `scope:` is for gateway usecases and not applicable for this generator. the reason why i put that in the PR description is to answer if asked the question "the TD documentation has (scope:) but its not part of the LDS resource name format of the generator". 

I also see why this is causing confusion. I've updated the PR description to reflect the same format that we are fixing this to.

ejona86 (approved):

ejona86:
Before merging, make sure to fix the commit description

main.go
arvindbr8: Gotcha! Thanks for explaining

easwars (approved):

PR #40, opened 2023-09-12
Enable xdstp style name by default for P2C and C2P cases

arvindbr8:
This change enables xdstp style names for P2C and C2P cases by default.


ejona86 (commented):
Why are we turning on includeXDSTPNameInLDS? I don't think we've tested that at all.

Note that your PR description mentions a test which you removed.

arvindbr8:
Per Mark's message: 

> Yes, we should add both the TD and C2P authorities by default. As a sanity check, the C2P authority should match the one being injected in the google-c2p resolver. The TD authority should set the client_listener_resource_name_template field to something of the form xdstp://traffic-director-global.xds.googleapis.com/envoy.config.listener.v3.Listener/<project_number>/<(network)|(mesh:mesh_name)|(scope:scope_name)>/id, as per[ IYKYK/td-xds-federation](http://goto.google.com/td-xds-federation).

I think we need to include both authorities with their corresponding LDS resource name templates.

arvindbr8:
Had a discussion offline with @ejona86. We have never tested using the new xdstp style name for regular TD authority. Hence we can only enable the C2P one in this release. 

however we need to make some changes before this can be merged. Blocked on #44

ejona86 (approved):

PR #45, opened 2023-09-21
Update cloudbuild-artificats.yaml build step

arvindbr8:
Since there are 2 packages now we need to use `go build .` to create the binary. `go test` still need to run for all the sub directories

### DIFF
--- a/tools/cloudbuild-artifacts.yaml
+++ b/tools/cloudbuild-artifacts.yaml
@@ -14,7 +14,7 @@
 
 steps:
 - name: golang:1.18
-  args: ['go', 'build', './...']
+  args: ['go', 'build', '.']
 - name: golang:1.18
   args: ['go', 'test', './...']
 - name: alpine


tools/cloudbuild-artifacts.yaml
```diff
@@ -14,7 +14,7 @@
 
 steps:
 - name: golang:1.18
-  args: ['go', 'build', './...']
+  args: ['go', 'build', '.']
```
easwars, line 17: Given that this is a go module, would `go build td-grpc-bootstrap/...` better? Same with `go test td-grpc-bootstrap/...`.  

It is interesting though that our module name is simply `td-grpc-bootstrap` and not something like `google.golang.org/td-grpc-bootstrap`. @ejona86 : Do you remember the history here?

arvindbr8: hmm @easwars .. `go build td-grpc-bootstrap/...` still doesnt create the td-grpc-bootstrap executable

also is there an advantage of using `go test td-grpc-bootstrap/...` vs `go test ./...` in this case?

easwars: > hmm @easwars .. `go build td-grpc-bootstrap/...` still doesnt create the td-grpc-bootstrap executable

Interesting. This is from `go help build`:
```
When compiling a single main package, build writes
the resulting executable to an output file named after
the first source file ('go build ed.go rx.go' writes 'ed' or 'ed.exe')
or the source code directory ('go build unix/sam' writes 'sam' or 'sam.exe').
The '.exe' suffix is added when writing a Windows executable.

When compiling multiple packages or a single non-main package,
build compiles the packages but discards the resulting object,
serving only as a check that the packages can be built.

The -o flag forces build to write the resulting executable or object
to the named output file or directory, instead of the default behavior described
in the last two paragraphs. If the named output is an existing directory or
ends with a slash or backslash, then any resulting executables
will be written to that directory.
```

easwars: > also is there an advantage of using `go test td-grpc-bootstrap/...` vs `go test ./...` in this case?

I tried to find if there is any difference between the two. No luck.

easwars (approved):

PR #48, opened 2023-11-03
remove experimental flags for on-by-default features

arvindbr8:
These features have been on-by-default for about 2 years. Lets get rid of some of the branches.

### DIFF
--- a/main.go
+++ b/main.go
@@ -29,8 +29,9 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/google/uuid"
 	"td-grpc-bootstrap/csmnamer"
+
+	"github.com/google/uuid"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


ejona86 (approved):
This does have a behavior change by removing  TRAFFICDIRECTOR_NETWORK_NAME/TRAFFICDIRECTOR_GCP_PROJECT_NUMBER . We'll need to pay attention to the interop tests.

main.go
```diff
@@ -40,8 +41,6 @@ var (
 	vpcNetworkName             = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone               = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
 	ignoreResourceDeletion     = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is not currently the case. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
-	includeV3Features          = flag.Bool("include-v3-features-experimental", true, "whether or not to generate configs which works with the xDS v3 implementation in TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
sergiitk, line 43: We still need this because `include-v3-features-experimental=False` enables xDS v2, which is still technically supported.

main.go
```diff
@@ -40,8 +41,6 @@ var (
...
-	includePSMSecurity         = flag.Bool("include-psm-security-experimental", true, "whether or not to generate config required for PSM security. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
```
sergiitk, line 44: @ejona86 do you think we'd need to keep an option to disable psm security fields?

main.go
ejona86: v2 works with this config, if the client only supports v2.

main.go
ejona86: Probably not. They are TLS and not bearer tokens, so can't really be abused in many ways. If the control plane doesn't reference them they are unused. I'm fine with adding such a flag if/when it is needed.

main.go
ejona86: We spoke offline. We have some tests that test v2-and-v3-supporting clients work when v3 is not in the bootstrap file. That should just use an old bootstrap generator (from before we changed the default), as that's the only situation a customer would be in that situation. We'll freeze the bootstrap generator it is using for v2 until the v2 testing stuff is deleted outright. @sergiitk will make that change internally; it'll be pretty easy.

sergiitk: ref cl/579998281

PR #49, opened 2023-11-09
add --gke-location-experimental to generate mesh ID for both regional and zonal GKE clusters

### LINKED ISSUE #47: Generate Mesh ID for zonal/regional clusters
For Proxyless Service Mesh usecase, the GKE cluster could be created to be either zonal or regional. Currently we only retrieve the zone from the metadata server. Which means using [`--generate-mesh-id-experimental`](github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/blob/master/main.go#L52) would generate an incorrect mesh ID for a mesh deployed in a regional cluster. 

The bootstrap generator is typically run in an init container for GKE workloads. We should find a way to figure out whether the deployment type of the cluster and we think the metadata server should be able to give us this information. 

Currently the [`getZone`](https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/blob/master/main.go#L356) method uses `http://metadata.google.internal/computeMetadata/v1/instance/zone` -- which retrieves the zone where the pod is located. 

Potential metadata server APIs that could work for us:

1.   `http://metadata.google.internal/computeMetadata/v1/instance/attributes/cluster-location`
    + which retrieves the location -- (zone/region)
2.   `http://metadata.google.internal/computeMetadata/v1/instance/attributes/kube-env`
    + which retrieves all kube envs. We need to look for `ZONE:` 

Before we proceed, we need a confirmation from the folks who maintain the "metadata server" which of these options are more reliable.

arvindbr8:
Fixes #47 

The value of `--locality-zone` flag has been used to auto generate Mesh ID as cluster location. However usage of zone to always generate the mesh ID is not correct. The location of the GKE cluster could either be either zonal or regional based on the deployment. 

Hence we need to introduce a new flag  `--gke-location-experimental` to determine the location of the GKE cluster. The flag is Ignored if not used with `--generate-mesh-id-experimental`. If unset the location of the GKE cluster can be determined using the following GKE metadata server (MDS) API `http://metadata.google.internal/computeMetadata/v1/instance/attributes/cluster-location`. 

For existing PSM users of the generator version https://github.com/GoogleCloudPlatform/traffic-director-grpc-boostrap/releases/v1.15.0 could continue using the same version of the generator if:

-  the right GKE cluster location is passed to `--location-zone` flag
- (or)  used in a zonally available cluster (because the existing behavior pulls the zone from the MDS server of the node in which the generator is being executed) 

Customers would have to explicitly upgrade their version to include this change in order to get the new UX of determining the location of the GKE cluster using the MDS server. In which case, the value passed to `--location-zone` would be ignored and the new flag would be used instead.

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ import (
 	"path"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"td-grpc-bootstrap/csmnamer"
@@ -40,6 +41,7 @@ var (
 	gcpProjectNumber           = flag.Int64("gcp-project-number", 0, "the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName             = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone               = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
+	clusterLocality            = flag.String("cluster-locality-experimental", "", "the locality of the cluster to use, instead of retrieving it from the metadata server. When set, this locality value is used to generate mesh ID. The availability of a cluster can be deployed be zonal or regional. This flag is EXPERIMENTAL and may be changed or removed in a later release")
 	ignoreResourceDeletion     = flag.Bool("ignore-resource-deletion-experimental", false, "assume missing resources notify operators when using Traffic Director, as in gRFC A53. This is not currently the case. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	secretsDir                 = flag.String("secrets-dir", "/var/run/secrets/workload-spiffe-credentials", "path to a directory containing TLS certificates and keys required for PSM security")
 	includeDeploymentInfo      = flag.Bool("include-deployment-info-experimental", false, "whether or not to generate config which contains deployment related information. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
@@ -64,6 +66,7 @@ func main() {
 		"alias of node-metadata. This flag is EXPERIMENTAL and will be removed in a later release")
 
 	flag.Parse()
+
 	if *gcpProjectNumber == 0 {
 		var err error
 		*gcpProjectNumber, err = getProjectId()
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -72,60 +75,44 @@ func main() {
...
 	}
+
+	deploymentType := getDeploymentType()
+
 	// Retrieve zone from the metadata server only if not specified in args.
-	zone := *localityZone
-	if zone == "" {
-		zone, err = getZone()
+	if *localityZone == "" {
+		*localityZone, err = getZone()
```
ejona86, line 88: I'm not wild about modifying the flags; that's modifying global state. Keep this as it was? Ditto for gkeClusterName and elsewhere.

arvindbr8 (commented):
Sorry about the delay in response due to vacation. @gnossen -- i've addressed to all your comments. Could you please take a look?

main.go
```diff
@@ -40,6 +41,7 @@ var (
 	gcpProjectNumber           = flag.Int64("gcp-project-number", 0, "the gcp project number. If unknown, can be found via 'gcloud projects list'")
 	vpcNetworkName             = flag.String("vpc-network-name", "default", "VPC network name")
 	localityZone               = flag.String("locality-zone", "", "the locality zone to use, instead of retrieving it from the metadata server. Useful when not running on GCP and/or for testing")
+	clusterLocality            = flag.String("cluster-locality-experimental", "", "the locality of the cluster to use, instead of retrieving it from the metadata server. When set, this locality value is used to generate mesh ID. The availability of a cluster can be deployed be zonal or regional. This flag is EXPERIMENTAL and may be changed or removed in a later release")
```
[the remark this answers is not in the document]
arvindbr8: Done

main.go
```diff
@@ -140,21 +124,38 @@ func main() {
...
+		default:
+			fmt.Fprintf(os.Stderr, "Error: unknown deployment type \n")
+			os.Exit(1)
 		}
 	}
 
-	var meshId string
-	if *generateMeshId {
+	meshId := *configMesh
+	if meshId == "" && *generateMeshId {
```
gnossen, line 134: There should be an error if both `--config-mesh` and `--generate-mesh-id` are supplied, right?

main.go
```diff
@@ -72,52 +74,34 @@ func main() {
...
 	}
+
 	ip, err := getHostIp()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: failed to determine host's ip: %s\n", err)
-		ip = ""
 	}
+
 	// Retrieve zone from the metadata server only if not specified in args.
 	zone := *localityZone
```
gnossen, line 84: What about backward compatibility for the `--locality-zone` flag? That is, if `--locality-zone` is supplied but not `--gke-location`, we should use that as the location of the cluster.

arvindbr8: I was trying to discuss this in the updated PR description. 

IMO, 

1.  `--locality-zone` should not be overloaded. This flag is being used to specify the `zone` in all use cases of the generator, expect when used with `--generate-mesh-id-experimental`.
    - `--generate-mesh-id-experimental` is marked experimental we should in the clear to change the dependencies.
    -  Adding b/c for the flag would open us to the vulnerability of users specifying a "region" to `--locality-zone` leading to incorrect bootstrap config generation (for eg: Node information [here](https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/blob/master/main.go#L262)).

1.  Customers wanting the new UX of determining the location of the cluster from MDS would any how explicitly upgrade the version of the generator. In which case, its easy to modify both `image` and `args` when updating k8s Deployment spec.

main.go
arvindbr8: if both `--config-mesh` and `--generate-mesh-id` are supplied, the former will have precedence and will be used as meshID

gnossen: This seems counterintuiive, no? Why not simply reject this case?

main.go
gnossen: >  In which case, its easy to modify both image and args when updating k8s Deployment spec.

This may be true, but we should discuss with the customer before going that route.

main.go
arvindbr8: Makes sense. Done.

main.go
arvindbr8: Discussed offline, and there is an action item to bring this up in the next customer meeting. Keeping this unresolved to update with what we hear back from the customer.

arvindbr8:
@gnossen -- I've addressed your comments again, ptal

gnossen (approved):

arvindbr8:
@ejona - gentle ping :)

main.go
```diff
@@ -140,21 +124,43 @@ func main() {
 				"GCP-ZONE":    zone,
 				"INSTANCE-IP": ip,
 			}
+		default:
+			fmt.Fprintf(os.Stderr, "Error: unknown deployment type \n")
```
ejona86, line 128: `dType` looks to be missing in the error string

ejona86 (approved):

main.go
arvindbr8: `getDeploymentType()` returns a literal `deploymentTypeUnknown` for cases where the request to the metadata server errored or is the response from MDS contains an unknown "Server" value. 

a better way would be perhaps to return the error from `getDeploymentType()` and put that in the error message. I can do that as a follow up

main.go
```diff
@@ -140,21 +124,43 @@ func main() {
...
+		if meshId != "" {
+			fmt.Fprint(os.Stderr, "Error: --config-mesh-experimental flag cannot be specified while --generate-mesh-id-experimental is also set.\n")
+			os.Exit(1)
+		}
+
+		clusterLocality := *gkeLocation
+		if clusterLocality == "" {
+			clusterLocality, err = getClusterLocality()
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error: unable to determine cluster locality from the metadata server with error: %s\n", err)
```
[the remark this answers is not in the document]
arvindbr8: Thanks. I've updated this in the next revision.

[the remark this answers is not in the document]
ejona86: Now the error is:

> Error: unable to generate mesh id: failed to determine GKE cluster locality: failed to determine GKE cluster locality: %s

That seems... worse.

main.go
ejona86: Okay. Well, the error isn't at all clear what happened or what the user should do. And I thought if you added more details you would have noticed the trailing space before the newline.

main.go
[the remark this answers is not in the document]
arvindbr8: my bad. Let me send a follow up PR right away to fixing that.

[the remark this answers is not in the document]
arvindbr8: sent https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/pull/51 to fix this

main.go
arvindbr8: https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/pull/51 in flight to address this

PR #53, opened 2024-01-17
add generator commit hash to node metadata

arvindbr8:
For some of the features in flight, it would be nice to determine the version of the generator used by looking at the Node metadata that is being passed to TD. This change add a new value to Node.Metadata keyed with `TRAFFICDIRECTOR_GRPC_BOOTSTRAP_GENERATOR_SHA` and is plumbed the git commit ID at HEAD.

### DIFF
--- a/main.go
+++ b/main.go
@@ -164,17 +164,10 @@ func main() {
 		meshId = meshNamer.GenerateMeshId()
 	}
 
-	var gitCommitHash string
-	info, ok := debug.ReadBuildInfo()
-	if !ok {
-		fmt.Fprintln(os.Stderr, "Error: unable to determine git commit hash from build information embedded in the running binary")
+	gitCommitHash, err := getGitCommitId()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: unable to determine git commit ID: %s\n", err)
 		os.Exit(1)
-	}
-	for _, setting := range info.Settings {
-		if setting.Key == "vcs.revision" {
-			gitCommitHash = setting.Value
-			break
-		}
 	}
 
 	input := configInput{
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -163,6 +164,19 @@ func main() {
 		meshId = meshNamer.GenerateMeshId()
 	}
 
+	var gitCommitHash string
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		fmt.Fprintln(os.Stderr, "Error: unable to determine git commit hash from build information embedded in the running binary")
```
ejona86, line 170: Is this really a fatal error, especially since if `vcs.revision` key is missing it continues without error?

arvindbr8: Thanks for catching that. i've fixed that in the latest revision.

ejona86 (commented):
I'm not wild about _requiring_ the vcs information, but that is easy enough to change in the future.

main_test.go
```diff
@@ -16,17 +16,61 @@ package main
...
 	"github.com/google/uuid"
 )
 
+// Tests to see if the value generated for Node Metadata keyed with
+// `TRAFFICDIRECTOR_GRPC_BOOTSTRAP_GENERATOR_SHA` is the same as the git commit
+// ID of HEAD.
+func TestCommitShaIdInNodeMetadata(t *testing.T) {
+
+	// Build td-grpc-bootstrap binary.
+	cmd := exec.Command("go", "build", ".")
```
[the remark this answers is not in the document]
arvindbr8: :')  I see your point. Updating this in the next rev.

We discussed something similar a while ago about how to get test the generator for its default behavior with flags. One of the testing methods you suggested was to do something similar to running `go build` and executing the binary. I was thinking all of our tests should look like this. 

Not for this change tho:
Do you think maybe adding a build and generate test to the github workflow might be worth it? And compare that to a golden file?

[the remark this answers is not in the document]
ejona86: I don't think I suggested `go build`. I might have suggested running the test binary though, like done in [Go's CGI tests](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/net/http/cgi/integration_test.go).

main_test.go
```diff
@@ -20,13 +20,26 @@ import (
...
 
+func TestGetGitCommitId(t *testing.T) {
+	commitId, err := getGitCommitId()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	re := regexp.MustCompile(`^[a-f0-9]{40}$`)
+	if !re.MatchString(commitId) {
+		t.Fatal("getGitCommitId(): returned an invalid commit ID. Want commit ID to be a valid SHA1 hash.")
```
ejona86, line 39: Include the commitId in the message?

ejona86 (approved):

main_test.go
arvindbr8: Done

PR #62, opened 2024-09-04
add GKE-Location to `TRAFFIC_DIRECTOR_CLIENT_ENVIRONMENT`

arvindbr8:
There exists a entry in node metadata which contains deployment related info. In addition to cluster name, this PR adds cluster location as well to the map, when run on GKE. 

There is a behavior change with this PR. The generator exits with errCode 1 if calls to the GKE metadata server fails to fetch cluster name or location.  This is to ensure that both values must be present if the generator is run in a GKE deployment.


main_test.go
```diff
@@ -555,6 +555,95 @@ func TestGenerate(t *testing.T) {
   },
   "server_listener_resource_name_template": "grpc/server?xds.resource.listening_address=%s",
   "client_default_listener_resource_name_template": "xdstp://traffic-director-global.xds.googleapis.com/envoy.config.listener.v3.Listener/123456789012345/thedefault/%s"
+}`,
+		},
+		{
+			desc: "with cluster namea and location specified",
```
gnossen, line 561: Nit: `s/namea/name/`

gnossen (approved):
LGTM

main.go
```diff
@@ -175,6 +175,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	if cn, err := getClusterName(); err == nil {
+		if cl, err := getClusterLocality(); err == nil {
```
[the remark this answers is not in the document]
arvindbr8: Discussed offline: 
Passing just one value in the node.metadata does not add value. 

> Also, tests?

I did manual testing in a GKE cluster. 

There are already unit tests around the helper functions that are used to fetch the information.  There is also another  test for the checking if the nodeMetadata map passed in to generate works correctly.

main.go
```diff
@@ -175,6 +175,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	if cn, err := getClusterName(); err == nil {
```
gnossen, line 178: Do we want to warn if we're unable to get cluster name or locality name? I think the previous revision was doing that.

main.go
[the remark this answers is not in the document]
easwars: Discussed offline. Since `getClusterName` and `getClusterLocality` already have unit tests, I'm ok with this PR not having tests.

easwars (approved):

main.go
```diff
@@ -175,6 +175,13 @@ func main() {
...
+		if cl, err := getClusterLocality(); err == nil {
+			nodeMetadata["GKE_CLUSTER_NAME"] = cn
```
ejona86, line 180: Isn't this already available in deploymentInfo's `GKE-CLUSTER`? (I don't claim to understand the difference between zone and cluster-location.) Are we just not using that because we'll need to add cluster location?

main.go
arvindbr8: I think that's a nice thing to have. I've added it back

main.go
arvindbr8: yes, and `TRAFFIC_DIRECTOR_CLIENT_ENVIRONMENT` is currently optional field added to node metadata,  controlled by a flag. We would want this to be added by default when run on GKE.

ejona86: Well, it could be enabled by default. Do we know what is going on with it and why we are doing something different here?

arvindbr8: The person who owns the design is OOO until next week. The discussion is blocked until then

arvindbr8: Finalized the discussion in the doc. Using `TRAFFIC_DIRECTOR_CLIENT_ENVIRONMENT` metadata to add cluster info.

ejona86 (approved):

PR #63, opened 2024-09-09
Add trusted_xds_server server feature for TD

kannanjgithub:
In order to address use-cases where authority rewriting may not be acceptable from a security perspective, adding a new server feature to the bootstrap config. The server feature is  specfied via the server_features field described in [gRFC A30](https://github.com/grpc/proposal/blob/master/A30-xds-v3.md). The feature is the string trusted_xds_server. (Note that the name is intentionally fairly general, since it may be used to trigger other security-sensitive functionality in the future.)

### DIFF
--- a/main.go
+++ b/main.go
@@ -259,8 +259,8 @@ func generate(in configInput) ([]byte, error) {
 		ChannelCreds: []creds{{Type: "google_default"}},
 	}
 
-	// Set xds_v3 Server Features.
-	xdsServer.ServerFeatures = append(xdsServer.ServerFeatures, "xds_v3")
+	// Set xds_v3 and trusted Server Features.
+	xdsServer.ServerFeatures = append(xdsServer.ServerFeatures, "xds_v3", "trusted_xds_server")
 
 	if in.ignoreResourceDeletion {
 		xdsServer.ServerFeatures = append(xdsServer.ServerFeatures, "ignore_resource_deletion")
--- a/main.go
+++ b/main.go
@@ -52,6 +52,7 @@ var (
 	configMesh             = flag.String("config-mesh", "", "Dictates which Mesh resource to use.")
 	generateMeshId         = flag.Bool("generate-mesh-id", false, "When enabled, the CSM MeshID is generated. If config-mesh flag is specified, this flag would be ignored. Location and Cluster Name would be retrieved from the metadata server unless specified via gke-location and gke-cluster-name flags respectively.")
 	includeXDSTPNameInLDS  = flag.Bool("include-xdstp-name-in-lds-experimental", true, "whether or not to use xdstp style name for listener resource name template. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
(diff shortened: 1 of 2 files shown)


ejona86 (commented):
This needs to start behind a flag, and after we do interop testing it will be stabilized.

Also, include notes in the commit message/PR description as to why we are making this change (reference the appropriate gRFC).

kannanjgithub:
> This needs to start behind a flag, and after we do interop testing it will be stabilized.
> 
> Also, include notes in the commit message/PR description as to why we are making this change (reference the appropriate gRFC).

I didn't do it because the gRFC talks adding the guard elsewhere:
```
Use of the RDS `auto_host_rewrite` field will be guarded by the
`GRPC_EXPERIMENTAL_XDS_AUTHORITY_REWRITE` env var.  The env var guard
will be removed once the feature passes interop tests.
```
I assumed each language will follow the same env var protection mechanism. Do we still need it here? I'm ok with adding it though.

ejona86:
The gRFC doesn't cover the update to this bootstrap generator. It only goes as far as defining changes to the bootstrap format. This generator is for a specific control plane.

"Flag" here means an actual flag, like `flag.Bool` at the top of main.go. We'd name it with `-experimental` and change the default when things are ready. It's not a huge concern in this case, but generally we want to verify that the config is correct before we enable it, and that is done by testing with clients that use the new values.

kannanjgithub:
> The gRFC doesn't cover the update to this bootstrap generator. It only goes as far as defining changes to the bootstrap format. This generator is for a specific control plane.
> 
> "Flag" here means an actual flag, like `flag.Bool` at the top of main.go. We'd name it with `-experimental` and change the default when things are ready. It's not a huge concern in this case, but generally we want to verify that the config is correct before we enable it, and that is done by testing with clients that use the new values.

Done.

main.go
```diff
@@ -52,6 +52,7 @@ var (
 	configMesh             = flag.String("config-mesh", "", "Dictates which Mesh resource to use.")
 	generateMeshId         = flag.Bool("generate-mesh-id", false, "When enabled, the CSM MeshID is generated. If config-mesh flag is specified, this flag would be ignored. Location and Cluster Name would be retrieved from the metadata server unless specified via gke-location and gke-cluster-name flags respectively.")
 	includeXDSTPNameInLDS  = flag.Bool("include-xdstp-name-in-lds-experimental", true, "whether or not to use xdstp style name for listener resource name template. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	isTrustedXdsServer     = flag.Bool("is-trusted-xds-server", false, "Whether to include the server feature trusted_xds_server for TD.")
```
ejona86, line 55: Include `-experimental` in the name and have the "This flag is EXPERIMENTAL". Unless we're going to keep this flag long-term to allow disabling this behavior.

kannanjgithub: Missed that, done now.

PR #66, opened 2024-09-13
remove `include-xdstp-name-in-lds-experimental` flag

arvindbr8:
Remove a default to true flag. 

Also refactor code, and update tests. Github does a very poor job at rendering the test diffs. sad

### DIFF
--- a/main.go
+++ b/main.go
@@ -51,7 +51,6 @@ var (
 	gceVM                  = flag.String("gce-vm-experimental", "", "GCE VM name to use, instead of reading it from the metadata server. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	configMesh             = flag.String("config-mesh", "", "Dictates which Mesh resource to use.")
 	generateMeshId         = flag.Bool("generate-mesh-id", false, "When enabled, the CSM MeshID is generated. If config-mesh flag is specified, this flag would be ignored. Location and Cluster Name would be retrieved from the metadata server unless specified via gke-location and gke-cluster-name flags respectively.")
-	includeXDSTPNameInLDS  = flag.Bool("include-xdstp-name-in-lds-experimental", true, "whether or not to use xdstp style name for listener resource name template. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


arvindbr8:
@ejona86 - a friendly ping

main.go
```diff
@@ -269,77 +266,63 @@ func validate(in configInput) error {
...
+		ServerUri:      in.xdsServerUri,
+		ChannelCreds:   []creds{{Type: "google_default"}},
+		ServerFeatures: []string{"xds_v3"},
 	}
-
 	if in.ignoreResourceDeletion {
-		xdsServer.ServerFeatures = append(xdsServer.ServerFeatures, "ignore_resource_deletion")
+		xs.ServerFeatures = append(xs.ServerFeatures, "ignore_resource_deletion")
 	}
+  if in.isTrustedXdsServer {
```
ejona86, line 280: Fix indentation

main.go
```diff
@@ -269,77 +266,63 @@ func validate(in configInput) error {
 }
 
 func generate(in configInput) ([]byte, error) {
-	xdsServer := server{
-		ServerUri:    in.xdsServerUri,
-		ChannelCreds: []creds{{Type: "google_default"}},
-	}
+	c := &config{}
```
ejona86, line 269: This is harder to read and verify. I really can't easily tell what is in the config except by looking at the test results. I don't see the need to rewrite this function when removing the `includeXDSTPNameInLDS` if, when all we needed to do was inline its contents into the `config` literal.

main_test.go
```diff
@@ -608,7 +553,6 @@ func TestGenerate(t *testing.T) {
     "cluster": "cluster",
     "metadata": {
       "INSTANCE_IP": "10.9.8.7",
-      "TRAFFICDIRECTOR_DIRECTPATH_C2P_IPV6_CAPABLE": true,
```
ejona86, line 611: How are there these two (here and just below) changes to the tests? Isn't deleting client_default_listener_resource_name_template clearly a bug?

arvindbr8:
Apologies for the bad commit. I've reverted all the other refactors and only removed the flag in this diff. Please take a look

ejona86 (approved):

PR #72, opened 2025-02-04
minor: address go-staticcheck warnings in main_test.go

sergiitk:
- Use `http.StatusForbidden` instead of `403`
- Remove redundant return


main_test.go
arvindbr8, line 1: Looks good to me except there is just one more [here](https://github.com/sergiitk/traffic-director-grpc-bootstrap/blob/a4dba4dd3d856ffa1b1b32b21d4aefa7ee513b88/main_test.go#L689)

arvindbr8 (approved):

main_test.go
sergiitk: Yea, missed one, thanks.

PR #75, opened 2025-10-09
Add support for Regional Traffic Director

fanymanea:
Add the regional Traffic Director authority to the bootstrap when the `xds-server-region` (a new flag added in this change) is passed in.

### DIFF
--- a/main.go
+++ b/main.go
@@ -272,7 +272,7 @@ func validate(in configInput) error {
 		return fmt.Errorf("config-mesh may only contain letters, numbers, and '-'. It must begin with a letter and must not exceed 64 characters in length")
 	}
 
-	regionRe := regexp.MustCompile(`^[a-z]*-[a-z]*[0-9]*$`)
+	regionRe := regexp.MustCompile(`^[a-z]+-[a-z]+[0-9]+$`)
 	if in.xdsServerRegion != "" && !regionRe.MatchString(in.xdsServerRegion) {
 		return fmt.Errorf("invalid xDS server region: %s", in.xdsServerRegion)
 	}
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -269,6 +272,11 @@ func validate(in configInput) error {
 		return fmt.Errorf("config-mesh may only contain letters, numbers, and '-'. It must begin with a letter and must not exceed 64 characters in length")
 	}
 
+	regionRe := regexp.MustCompile(`^[a-z]*-[a-z]*[0-9]*$`)
```
ejona86: Should some of these `*` be `+` instead, to have at least one alpha character?

main.go
```diff
@@ -55,6 +55,7 @@ var (
 	generateMeshID             = flag.Bool("generate-mesh-id", false, "When enabled, the CSM MeshID is generated. If config-mesh flag is specified, this flag would be ignored. Location and Cluster Name would be retrieved from the metadata server unless specified via gke-location and gke-cluster-name flags respectively.")
 	includeAllowedGrpcServices = flag.Bool("include-allowed-grpc-services-experimental", false, "When enabled, generates `allowed_grpc_services` map that includes current xDS Server URI. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
 	isTrustedXDSServer         = flag.Bool("is-trusted-xds-server-experimental", false, "Whether to include the server feature trusted_xds_server for TD. This flag is EXPERIMENTAL and may be changed or removed in a later release.")
+	xdsServerRegion            = flag.String("xds-server-region", "", "The Cloud region (e.g. 'us-central1') of the xDS server to be used for regional setup.")
```
ejona86: We normally would start with an experimental name, and once testing passes we provide a non-experimental argument. Are there any testing plans for this?

ejona86 (commented):
This looks fine to me overall. Just a few minor discussions.

main.go
fanymanea: Done.

main.go
fanymanea: I did the experimental thing.
About testing, I asked around and I was pointed to go/grpc/td-grpc-bootstrap-release. I might be the one who adds the guitar tests for Regional TDOM, I'm not sure. I'm not working in this area usually, but I am helping out with this now. Anyway, my understanding is that this change needs to get in before we run those tests.

ejona86 (approved):

main.go
ejona86: I wrote go/grpc/td-grpc-bootstrap-release, so I'm familiar with that. We can continue discussion of what the testing approach looks like internally.

PR #77, opened 2025-12-04
Add support for Cloud Run

victor-mora:
Allow MDS to recognize Cloud Run deployments and add support for regional locality and Instance ID.

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,9 +33,9 @@ import (
 	"strings"
 	"time"
 
-	"google3/third_party/golang/github_com/GoogleCloudPlatform/traffic_director_grpc_bootstrap/v/v0/csmnamer/csmnamer"
+	"td-grpc-bootstrap/csmnamer"
 
-	"google3/third_party/golang/googleuuid/uuid"
+	"github.com/google/uuid"
 )
 
 var (
--- a/main.go
+++ b/main.go
@@ -91,15 +91,12 @@ func main() {
 		}
 	}
 
-	var ip string
-	var err error
-	ip, err = getHostIP()
+	ip, err := getHostIP()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: failed to determine host's IP: %s\n", err)
 	}
 
-	var dType deploymentType
-	dType, err = getDeploymentType()
+	dType, err := getDeploymentType()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: unable to determine deployment type: %s\n", err)
 	}
@@ -169,11 +166,11 @@ func main() {
 			"INSTANCE-IP": ip,
 		}
 	case deploymentTypeCloudRun:
-		instanceID, err := getInstanceID()
+		instanceID, err := getCloudRunInstanceID()
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
+			os.Exit(1)
 		} else {
-			fmt.Printf("instanceID: %s\n", instanceID)
 			deploymentInfo = map[string]string{
 				"INSTANCE-ID": instanceID,
 			}
@@ -492,7 +489,7 @@ func getVMName() string {
 	return string(vm)
 }
 
-func getInstanceID() (string, error) {
+func getCloudRunInstanceID() (string, error) {
 	instanceID, err := getFromMetadata("http://metadata.google.internal/computeMetadata/v1/instance/id")
 	if err != nil {
 		return "", fmt.Errorf("failed to determine instance id: %w", err)
--- a/main.go
+++ b/main.go
@@ -33,9 +33,9 @@ import (
 	"strings"
 	"time"
 
-	"td-grpc-bootstrap/csmnamer"
+	"google3/third_party/golang/github_com/GoogleCloudPlatform/traffic_director_grpc_bootstrap/v/v0/csmnamer/csmnamer"
 
-	"github.com/google/uuid"
+	"google3/third_party/golang/googleuuid/uuid"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -90,27 +91,41 @@ func main() {
 		}
 	}
 
-	ip, err := getHostIP()
+	var ip string
+	var err error
```
easwars: Why is this change required? 

`ip, err := getHostIP()` followed by `dType, err := getDeploymentType()` should work.

main.go
```diff
@@ -458,6 +492,14 @@ func getVMName() string {
 	return string(vm)
 }
 
+func getInstanceID() (string, error) {
```
easwars: If this is something specific to cloudRun, can we make the function name more descriptive? Like `getCloudRunInstanceID` maybe?

main.go
```diff
@@ -153,6 +168,16 @@ func main() {
 			"GCP-ZONE":    zone,
 			"INSTANCE-IP": ip,
 		}
+	case deploymentTypeCloudRun:
+		instanceID, err := getInstanceID()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
+		} else {
+			fmt.Printf("instanceID: %s\n", instanceID)
```
easwars: Was this added for debugging during development? If so, please remove.

main.go
```diff
@@ -153,6 +168,16 @@ func main() {
 			"GCP-ZONE":    zone,
 			"INSTANCE-IP": ip,
 		}
+	case deploymentTypeCloudRun:
+		instanceID, err := getInstanceID()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
```
easwars: Should the generator exit at this point? Is there any point in continuing without the instance-id?

main.go
```diff
@@ -308,7 +335,8 @@ func generate(in configInput) ([]byte, error) {
 			ID:      fmt.Sprintf("projects/%d/networks/%s/nodes/%s", in.gcpProjectNumber, networkIdentifier, uuid.New().String()),
 			Cluster: "cluster", // unused by TD
 			Locality: &locality{
-				Zone: in.zone,
+				Zone:   in.zone,
+				Region: in.region,
```
easwars: This would cause non-cloudRun deployments that previously emitted a configuration where region was not specified to change to a configuration with an empty region. I'm not sure if we want that.

easwars:
@victor-mora : It looks like there is an open design proposal (authored by you) that is still in review. We should get that approved before we can make progress with this PR.

main.go
victor-mora: Changed to `ip, err := getHostIP()` followed by `dType, err := getDeploymentType()` in latest commit: https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap/pull/77/commits/33fb210018abef0dbb2691925516d783ce294f5c

main.go
victor-mora: Changed to `getCloudRunInstanceID`.

main.go
victor-mora: Left the `fmt.Fprintf(os.Stderr, "Warning: %s\n", err)` to match the other metadata calls, but removed `fmt.Printf("instanceID: %s\n", instanceID)`.

main.go
victor-mora: Added `os.Exit(1)`.

main.go
victor-mora: Is that the case? The `locality` struct in `main.go` is:
```
type locality struct {
	Region  string `json:"region,omitempty"`
	Zone    string `json:"zone,omitempty"`
	SubZone string `json:"sub_zone,omitempty"`
}
```
With `omitempty`, won't empty string entries be omitted from the resulting json?

main.go
```diff
@@ -458,6 +468,14 @@ func getVMName() string {
 	return string(vm)
 }
 
+func getCloudRunInstanceID() (string, error) {
+	instanceID, err := getFromMetadata("http://metadata.google.internal/computeMetadata/v1/instance/id")
+	if err != nil {
+		return "", fmt.Errorf("failed to determine instance id: %w", err)
```
easwars: Is there a reason to wrap the error using the `%w` verb here?

main.go
```diff
@@ -153,6 +153,16 @@ func main() {
 			"GCP-ZONE":    zone,
 			"INSTANCE-IP": ip,
 		}
+	case deploymentTypeCloudRun:
+		instanceID, err := getCloudRunInstanceID()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
```
easwars: Since we are killing the executable in the next line, it might be better to call this out as "error" or "fatal" instead of "warning".

main.go
```diff
@@ -458,6 +468,14 @@ func getVMName() string {
 	return string(vm)
 }
 
+func getCloudRunInstanceID() (string, error) {
```
easwars, line 471: We should have a unit test for this function, just like how we have one for `getClusterLocality`.

main_test.go
```diff
@@ -743,6 +743,84 @@ func TestGenerate(t *testing.T) {
   },
   "server_listener_resource_name_template": "grpc/server?xds.resource.listening_address=%s",
   "client_default_listener_resource_name_template": "xdstp://traffic-director-global.xds.googleapis.com/envoy.config.listener.v3.Listener/123456789012345/thedefault/%s"
+}`,
+		},
+    {
+			desc: "happy case for Cloud Run deployment",
```
easwars: Since the only difference for the cloudrun environment is how the deployment information is generated, I believe this test is not useful. Here, we are populating the `deploymentInfo`. As long as we test `getCloudRunInstanceID`, we should be good without this test. What do you think?

main.go
victor-mora: Added unit test for `getCloudRunInstanceID`

main.go
victor-mora: Changed to `%v` for consistency with other functions.

main.go
victor-mora: Changed to "Error".

main_test.go
victor-mora: Removed this test in favor of the `getCloudRunInstanceID` unit test.

main_test.go
```diff
@@ -881,6 +881,52 @@ func TestGetClusterLocality(t *testing.T) {
...
+				if r.Header.Get("Metadata-Flavor") != "Google" {
+					http.Error(w, "Missing Metadata-Flavor", http.StatusForbidden)
+					return
+				}
+				w.Write([]byte("instance-id-123"))
+			},
+			want: "instance-id-123",
+		},
+		{
+			desc: "no_response_from_server",
```
easwars: Nit: Can we use a better name for this. The current name looks like the request times out because of no response from the server. "server_returns_403" or something like that.

easwars (approved):
LGTM, modulo one minor nit

easwars:
@ejona86 : In case, you want to take a look.

main.go
```diff
@@ -153,6 +153,16 @@ func main() {
 			"GCP-ZONE":    zone,
 			"INSTANCE-IP": ip,
 		}
+	case deploymentTypeCloudRun:
+		instanceID, err := getCloudRunInstanceID()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error: %s\n", err)
```
ejona86: We should add more context to this message. Maybe "Error: generating cloud run deployment info: %s\n"

ejona86 (approved):
One small thing, otherwise looks good.

main_test.go
victor-mora: Changed name to `server_returns_403`.

main.go
victor-mora: Added context to error message.

PR #82, opened 2026-07-14
Fix regional authority ServerURI generation to respect --xds-server-uri overrides

shivaspeaks:
cl/947945886 has some context on staging failure that was seen.

When --xds-server-region is used alongside --xds-server-uri (e.g., when targeting staging sandboxes), the generator previously hardcoded the regional authority's ServerURI to production (trafficdirector.<region>.rep.googleapis.com:443). This caused gRPC clients attempting regional xDS lookups to connect to production instead of the intended test endpoint.

This fix checks whether the provided --xds-server-uri already contains the target region. If it does (as is the case with staging sandbox URIs), it applies that custom URI directly to the regional authority entry (traffic-director.<region>.xds.googleapis.com). Otherwise, it falls back to generating the standard production regional endpoint, ensuring staging tests route to the correct control plane while preserving existing testing and production behavior.

### DIFF
--- a/main.go
+++ b/main.go
@@ -338,7 +338,9 @@ func generate(in configInput) ([]byte, error) {
 	if in.xdsServerRegion != "" {
 		regionalTdAuthority := fmt.Sprintf("traffic-director.%s.xds.googleapis.com", in.xdsServerRegion)
 		regionalXdsServer := xdsServer
-		regionalXdsServer.ServerURI = fmt.Sprintf("trafficdirector.%s.rep.googleapis.com:443", in.xdsServerRegion)
+		if !strings.Contains(regionalXdsServer.ServerURI, in.xdsServerRegion) {
+			regionalXdsServer.ServerURI = fmt.Sprintf("trafficdirector.%s.rep.googleapis.com:443", in.xdsServerRegion)
+		}
 
 		c.Authorities[regionalTdAuthority] = Authority{
 			XDSServers: []server{regionalXdsServer},
--- a/main_test.go
+++ b/main_test.go
@@ -263,6 +263,95 @@ func TestGenerate(t *testing.T) {
 }`,
 		},
 		{
+			desc: "happy case with staging regional config setup",
+			input: configInput{
+				xdsServerURI:     "staging-trafficdirector.antarctica-central1.rep.sandbox.googleapis.com:443",
+				gcpProjectNumber: 123,
+				vpcNetworkName:   "test",
+				ip:               "10.9.8.7",
+				zone:             "antarctica-central1",
+				gitCommitHash:    "12345",
+				xdsServerRegion:  "antarctica-central1",
+			},
+			wantOutput: `{
+  "xds_servers": [
+    {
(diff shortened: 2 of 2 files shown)


shivaspeaks:
CC: @sergiitk

shivaspeaks:
**Note:** 
The comparison in.xdsServerURI != *xdsServerURI always evaluated to false because in.xdsServerURI is initialized to *xdsServerURI in main(). To fix this, we should compare in.xdsServerURI against the default value of the flag ("trafficdirector.googleapis.com:443"). Added a commit.

shivaspeaks:
When the test was originally written, the bug still existed in the code. Because the code was blindly overwriting the URI, the test was written to expect the output to be "trafficdirector.antarctica-central1.rep.googleapis.com:443".

main.go
```diff
@@ -350,8 +350,12 @@ func generate(in configInput) ([]byte, error) {
 	if in.xdsServerRegion != "" {
 		regionalTdAuthority := fmt.Sprintf("traffic-director.%s.xds.googleapis.com", in.xdsServerRegion)
 		regionalXdsServer := xdsServer
-		regionalXdsServer.ServerURI = fmt.Sprintf("trafficdirector.%s.rep.googleapis.com:443", in.xdsServerRegion)
-
+		if in.xdsServerURI != "trafficdirector.googleapis.com:443" {
```
ejona86, line 353: generate() doesn't depend on the args, and this is not at all clear what it is doing. If we were going to do it this way we would have make the td uri a constant and referenced it here.

But I think we need a new configInput field so that main() can compute the configuration. It'd be fine to have a field that is an override; it'd only be used when non-empty.

main_test.go
```diff
@@ -222,7 +222,7 @@ func TestGenerate(t *testing.T) {
     "traffic-director.antarctica-central1.xds.googleapis.com": {
       "xds_servers": [
         {
-          "server_uri": "trafficdirector.antarctica-central1.rep.googleapis.com:443",
+          "server_uri": "example.com:443",
```
ejona86, line 225: This is the only test that includes the regional server_uri. We need a test verifying the td regional string.

ejona86 (commented):
Submitting what I was writing, but I see you just pushed a new version.

ejona86 (commented):
Honestly, I'm not sure what we are testing by running regional tests in staging, if it is just going to use the same instance control plane instance. I don't think regional support makes much sense in a testing environment; it has no meaning.

ejona86:
> I don't think regional support makes much sense in a testing environment; it has no meaning.

Ehh... I guess there is configuration like ClientListenerResourceNameTemplate which would be tested, sorta. Okay.

shivaspeaks:
Apologies, I messed up with the branch and did the force push.
> You completely defeated the test that verifies that regional uses its own URI. And "randomly" comparing against the TD uri to see if the original flag changed is pretty hacky in that part of the code.

I changed the the logic to only overwrite the regional URI if it does not already contain the target region. So now regional URIs (e.g., staging sandbox endpoints) will now pass through to the regional authority untouched.

